Issue #48, opened 2015-01-02
Network_interface.Internal_error("Not_found") - /proc/net/bonding/


bugsyb:
For some reasons running 
xe pif-scan host-uuid=....

it looks for /proc/net/bonding/<bridge_interface_name> whilst it does not exist on normal system. Tested on cleanly installed system too.
System in question: Ubuntu 14.04 LTS, Xen 4.4 (amd64).
I'm guessing it might be related to:
https://github.com/xapi-project/xcp-networkd/issues/38

bridge setup in standard Linux way using brctl and properly visible at Xen level.

xe network-list
uuid ( RO)                : 5369a468-f639-b7ec-ddc1-b1d5861da84e
          name-label ( RW): Pool-wide network associated with em1
    name-description ( RW):
              bridge ( RO): brem1

em1 is standard ethernet interface.

xe pif-list
uuid ( RO)                  : b6a218d3-8f7b-f9cd-c1f1-3f16a63cc785
                device ( RO): em1
    currently-attached ( RO): true
                  VLAN ( RO): -1
          network-uuid ( RO): 5369a468-f639-b7ec-ddc1-b1d5861da84e

trying to scan for interfaces:
xe pif-scan host-uuid=0d83e699-df0b-40db-9934-b5cbc9e5eafe
The server failed to handle your request, due to an internal error.  The given message may give details useful for debugging the problem.

Syslog messages:
Jan  2 14:06:09 silver xcp-networkd: [error|silver|174|PIF.scan R:01e4a6892f20|network_utils] Error: could not read /proc/net/bonding/brem1.

Jan  2 14:06:09 silver xapi: [debug|silver|1989 UNIX /var/lib/xcp/xapi||cli] Xapi_cli.exception_handler: Got exception INTERNAL_ERROR: [ Network_interface.Internal_error("Not_found") ]

Full dump:
Jan  2 14:06:09 silver xapi: [ info|silver|1989 UNIX /var/lib/xcp/xapi||cli] xe pif-scan host-uuid=0d83e699-df0b-40db-9934-b5cbc9e5eafe username=root password=(omitted)
[...]

PR #49, opened 2015-01-08
CA-151464: Don't assume interface is bonded if master sysfs key exists

simonjbeaumont:
In 3.12+ kernels, this key also exists for OVS ports.

Signed-off-by: Si Beaumont <simon.beaumont@citrix.com>

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -456,8 +456,13 @@ module Linux_bonding = struct
 
 	let get_bond_master_of slave =
 		try
-			let path = Unix.readlink (Sysfs.getpath slave "master") in
-			Some (List.hd (List.rev (String.split '/' path)))
+			let master_symlink = Sysfs.getpath slave "master" in
+			let master_path = Unix.readlink master_symlink in
+			let slaves_path = Filename.concat master_symlink "bonding/slaves" in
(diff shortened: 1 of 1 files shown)


lib/network_utils.ml
```diff
@@ -456,8 +456,13 @@ module Linux_bonding = struct
 
 	let get_bond_master_of slave =
 		try
-			let path = Unix.readlink (Sysfs.getpath slave "master") in
-			Some (List.hd (List.rev (String.split '/' path)))
+			let master_symlink = Sysfs.getpath slave "master" in
+			let master_path = Unix.readlink master_symlink in
+			let slaves_path = Filename.concat master_symlink "bonding/slaves" in
+			let slaves = Sysfs.read_one_line slaves_path |> String.split ' ' in
```
robhoes: Could it be enough to check for the existence of `slaves_path` to conclude that this is indeed a bond slave? I think that if this path exists, we can rely on `master_path` to be what we want.

simonjbeaumont:
@robhoes is this better?

simonjbeaumont:
Fixes #48.

robhoes:
Yes, looks good. I'm happy if xen-git is.

PR #52, opened 2015-05-20
CP-12093: Add dcbtool to filter the NICs supporting FCoE

sharady:


networkd/network_server.ml
```diff
@@ -317,6 +317,14 @@ module Interface = struct
 			Ethtool.set_offload name params
 		) ()
 
+	let is_fcoe_supported _ dbg ~name =
+		Debug.with_thread_associated dbg (fun () ->
+			let output = Dcbtool.is_fcoe_supported name in
+			if String.has_substr output "Successful"
```
euanh: Why don't you do this part in Dcbtool.is_fcoe_supported, making that function return a bool instead of a string?

sharady: Make sense, I am updating the pull request. Thanks

lib/network_utils.ml
```diff
@@ -546,6 +547,17 @@ module Dhclient = struct
 		end
 end
 
+module Dcbtool = struct
+	let call ?(log=false) args =
+		call_script ~log_successful_output:log dcbtool args
+
+	let is_fcoe_supported name =
+		let output = call ["gc"; name; "dcb"] in
```
robhoes: I don't think we can assume that `dcbtool` always present, so we need to add some error handling here.

lib/network_utils.ml
```diff
@@ -33,6 +33,7 @@ let modprobe = "/sbin/modprobe"
 let ethtool = ref "/sbin/ethtool"
 let bonding_dir = "/proc/net/bonding/"
 let dhcp6c = "/sbin/dhcp6c"
+let fcoedriver = "/opt/xensource/fcoe/fcoe_driver.py"
```
robhoes: It would be good to allow setting this from the xcp-networkd config file. You can follow the way it is done for `ethtool`:
* Make the string above a `ref`.
* Add a section to networkd.ml, like this: https://github.com/xapi-project/xcp-networkd/blob/master/networkd/networkd.ml#L37

robhoes:
It looks fine otherwise, but please squash the patches.

sharady:
Thanks @robhoes , for the review.
I have squashed the commits.

networkd/networkd.ml
```diff
@@ -39,6 +39,12 @@ let resources = [
     essential = true;
     path = Network_utils.ethtool;
     perms = [ Unix.X_OK ];
+  };
+  { Xcp_service.name = "fcoedriver";
+    description = "used to identify fcoe interfaces";
+    essential = true;
```
robhoes: I'm going to set `essential` to `false` before merging this.

robhoes:
I've merged this (even though github thinks I've not).

PR #53, opened 2015-06-11
CP-9895: Added originator for xapi login_with_password

koushikcgit:
The login_with_password XenAPI call takes an "originator"
string as its fourth parameter.

If a client does this, then it gets its own pool of xapi sessions.
Moreover it will not have its xapi sessions destroyed prematurely
as a result of some other misbehaving client that keeps creating
sessions and not logging out of them.

This patch adds the "originator" to all invokes of
login_with_password.


xen-git:
Can one of the admins verify this patch for testing?

koushikcgit:
This PR depends on https://github.com/xapi-project/xen-api-client/pull/28

thomassa:
You can add the version string from the Version module:
`~originator:("xcp-networkd v" ^ Version.version)`

thomassa:
@xen-git Test this please

thomassa:
In it goes!

PR #54, opened 2015-06-17
CA-113824: Refactor code for getting devices

ravippandey:


robhoes:
And as a bonus, could you add an example config file, including all config options that are currently supported (commented out)?

networkd/network_monitor_thread.ml
```diff
@@ -98,6 +106,13 @@ let get_link_stats () =
 
 	let cache = Link.cache_alloc s in
 	let links = Link.cache_to_list cache in
+	let links = List.filter (fun link ->
+		let name = standardise_name (Link.get_name link) in
```
robhoes: Hmm... the above is now called twice (here and in L117). Could you factor it out using List.map?

robhoes:
Looks good now, thanks!

PR #56, opened 2015-07-07
CP-12093: Add fcoe_driver interface to identify FCoE supported NICs

sharady:


sharady:
As suggested by @robhoes on slack, fcoe changes should go via trunk-ring3.
Creating PR on master.

robhoes:
This was earlier reviewed on the "fcoe" branch, and tested there. This is now good to go into master.

PR #58, opened 2015-11-04
Start xcp-networkd with priority 14

robhoes:
This ensures that xcp-networkd is started after forkexecd, which has a start
priority of 13 (since 97cc4745 in forkexecd.git).

### DIFF
--- a/scripts/init.d-networkd
+++ b/scripts/init.d-networkd
@@ -2,7 +2,7 @@
 #
 # xcp-networkd     Start/Stop the XCP networking daemon
 #
-# chkconfig: 2345 13 76
+# chkconfig: 2345 14 76
 # description: XCP networking daemon
 # processname: xcp-networkd
 # pidfile: /var/run/xcp-networkd.pid


johnelse:
Is this actually the script that gets installed on a XenServer? Or is it [this one](https://github.com/xenserver/xen-api-libs-specs/blob/master/SOURCES/xcp-networkd-init)?

robhoes:
Ehm... probably the other one, so I'll raise another PR :)
Might as well merge this one too, to keep them in sync?

thomassa:
Is there any way we could have just one of them?
If not, we should probably have a comment in each as a reminder about the other one.

robhoes:
It would be great to have just one of them. I don't know why there is a separate one in xen-api-libs-specs... Perhaps @jonludlam knows why?

johnelse:
I suspect this one just got put here for safe keeping when we split up xen-api - I'm pretty sure the one in xen-api-libs-specs is the only one that ends up in the product.

euanh:
@johnelse is right, the spec file in xen-api-libs-specs refers to the xcp-networkd-init in that repository.

robhoes:
So what is the right thing to do: keep and update the script in this repo, or remove it?

johnelse:
I'm all for removing this one, assuming it's the same as the one in xen-api-libs-specs.

robhoes:
Funny enough, even xen-api-libs-specs has two versions of the init script... (see xenserver/xen-api-libs-specs#131).

johnelse:
Of course :wink:

robhoes:
I'll close this PR, and open a new one to remove the init script.

PR #59, opened 2015-11-04
Remove init script

robhoes:
There is one (two?) in xen-api-libs-specs already, which is confusing.


robhoes:
Also see #58.

johnelse:
:boom:

PR #60, opened 2015-11-10
CA-184010: Call fcoedriver helper program with timeout=5seconds

sharady:
Xapi startup sequence should not get blocked if fcoedriver fails to
fetch the fcoe capable status.
We must timeout the fcoedriver, if it's not returning.

Signed-off-by: sharad yadav <sharad.yadav@citrix.com>


sharady:
Dependent on https://github.com/xapi-project/forkexecd/pull/11

thomassa:
Closing this for the time being; can re-open after update.

robhoes:
Perhaps just `?timeout` as the param, which gives you `None` if `~timeout` is absent from the function call.

PR #61, opened 2015-11-12
CA-184010: Add a timeout to script invocations

jonludlam:
Signed-off-by: Jon Ludlam <jonathan.ludlam@citrix.com>


robhoes:
Needs https://github.com/xapi-project/forkexecd/pull/12

johnelse:
Looks good!

robhoes:
Hold on... 60s would not be enough for dhclient... We should give that call a longer timeout, because it has a timeout built in it self.

PR #62, opened 2015-12-02
Don't use a timeout when executing dhclient

robhoes:
Dhclient has a built-in timeout of its own, after which is backgrounds itself
and unblocks the caller.  We don't want dhclient to get killed by networkd
after some arbitrary timeout.

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -35,13 +35,13 @@ let bonding_dir = "/proc/net/bonding/"
 let dhcp6c = "/sbin/dhcp6c"
 let fcoedriver = ref "/opt/xensource/libexec/fcoe_driver"
 
-let call_script ?(log_successful_output=false) ?(timeout=60.0) script args =
+let call_script ?(log_successful_output=false) ?(timeout=Some 60.0) script args =
 	try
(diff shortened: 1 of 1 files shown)


lib/network_utils.ml
```diff
@@ -35,13 +35,17 @@ let bonding_dir = "/proc/net/bonding/"
...
 	try
 		Unix.access script [ Unix.X_OK ];
 		(* Use the same $PATH as xapi *)
 		let env = [| "PATH=" ^ (Sys.getenv "PATH") |] in
 		info "%s %s" script (String.concat " " args);
-		let (out,err) = Forkhelpers.execute_command_get_output ~env ~timeout script args in
+		let (out,err) =
+			match timeout with
+			| Some t -> Forkhelpers.execute_command_get_output ~env ~timeout:t script args
+			| None -> Forkhelpers.execute_command_get_output ~env script args
```
johnelse: I think this whole match can be simplified to

```
Forkhelpers.execute_command_get_output ~env ?timeout script args
```

robhoes: You are probably right :)

johnelse:
Looks good!

PR #63, opened 2015-12-04
CA-192116: On startup, each operation must be best effort

robhoes:
This means that if we cannot setup one of the bridges, for some reason, we must
still try to configure the remaining ones.


johnelse:
Looks good!

PR #64, opened 2016-01-05
CP-15312: Remove harmless exception from toolstack startup

koushikcgit:

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -992,8 +992,12 @@ module Dhcp6c = struct
 		Printf.sprintf "/var/run/dhcp6c-%s.pid" interface
 
 	let start interface =
-		ignore (call_script dhcp6c [interface])
+		try
+			ignore (call_script dhcp6c [interface])
+		with _ -> ()
 
 	let stop interface =
-		ignore (call_script dhcp6c ["-r"; "all"; interface])
+		try
+			ignore (call_script dhcp6c ["-r"; "all"; interface])
+		with _ -> ()
 end


robhoes:
The `Network_utils` module is meant to be a library of lower-level Linux networking functions and should not just ignore exceptions. The higher-level `Network_server` module uses `Network_utils` and may choose to catch and discard exceptions.

In fact, the `Interface.make_config` function calls `set_ipv6_conf` (which calls `Dhcp6.start` and `Dhcp6.stop` from `Network_utils`) and drops the exception already. So what exactly is this patch doing?

koushikcgit:
@robhoes : Even though the exceptions were being dropped, the xensource.log still had the raise exception logged which is what needed to be removed. I was trying to do that in this patch.

robhoes:
Unfortunately, `call_script` logs the error even if you catch the exception.

robhoes:
After a bit of looking around, I conclude that errors are logged from `set_ipv6_conf` in two places:
- By `call_script`, which reports that it cannot find the `dhcp6c` program.
- By `Debug.with_thread_associated`, which logs the backtrace.

Koushik's patch avoids the latter. However, I'm still not happy with the approach. Much better would be to actually remove the cause of the exception. It turns out that all uses of dhcp6c can now be replaced by calls to dhclient (using the `-6` option). The `Dhclient` module in network_utils.ml already has code for this. I therefore propose to fix the issue by switching over to `Dhclient` and drop the `Dhcp6c` module, in a new PR.

robhoes:
I'll close this PR based on the above.

PR #65, opened 2016-01-11
CP-15312: Remove dhcp6 exception from toolstack startup

koushikcgit:
dhclient handles all ipv6 related configurations and hence dhcp6
is not required

### DIFF
--- a/networkd/network_server.ml
+++ b/networkd/network_server.ml
@@ -193,31 +193,36 @@ module Interface = struct
 			match conf with
 			| None6 ->
 				if List.mem name (Sysfs.list ()) then begin
-					Dhcp6c.stop name;
+					if Dhclient.is_running ~ipv6:true name then
+						ignore (Dhclient.stop ~ipv6:true name);
 					Sysctl.set_ipv6_autoconf name false;
 					Ip.flush_ip_addr ~ipv6:true name
 				end
 			| Linklocal6 ->
 				if List.mem name (Sysfs.list ()) then begin
-					Dhcp6c.stop name;
+					if Dhclient.is_running ~ipv6:true name then
+						ignore (Dhclient.stop ~ipv6:true name);
 					Sysctl.set_ipv6_autoconf name false;
 					Ip.flush_ip_addr ~ipv6:true name;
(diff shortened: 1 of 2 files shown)


koushikcgit:
Previous work: https://github.com/xapi-project/xcp-networkd/pull/64

koushikcgit:
Adding @robhoes  as reviewer

robhoes:
@koushikcgit The `Dhclient` module has some functions to to check whether a `dhclient` process is already running, and doing the right thing based on that. Please have a look at the IPv4 case, and do something similar. This will make the high-level function idempotent and should avoid unnecessary errors in the logs.

robhoes:
Travis is again complaining about `Wrong type NilClass. Not allowed!` :(

jonludlam:
I think this might have been because gandalf was full. I have performed a log purge (including a console log from an errant mirage VM that was spewing rubbish into /tmp) so it might work now.

koushikcgit:
@robhoes: Updated

networkd/network_server.ml
```diff
@@ -193,31 +193,36 @@ module Interface = struct
...
 				end
 			| DHCP6 ->
-				Dhcp6c.stop name;
+				if Dhclient.is_running ~ipv6:true name then
+					ignore (Dhclient.stop ~ipv6:true name);
 				Sysctl.set_ipv6_autoconf name false;
 				Ip.flush_ip_addr ~ipv6:true name;
 				Ip.set_ipv6_link_local_addr name;
-				Dhcp6c.start name
+				ignore (Dhclient.start ~ipv6:true name [])
```
robhoes: Please use `Dhclient.ensure_running` here like in the v4 case. This avoids unnecessarily stopping dhclient.

koushikcgit: The `ensure_running` stops and restarts the Dhclient, while here, I need to perform some steps before starting.
Hence keeping it as stop followed by start.

robhoes: Ok, I agree that it is safer to keep it this way.

robhoes:
:shipit: after the :icecream:

PR #66, opened 2016-02-05
CA-149863: Take bond mode into account when determining bond speed

robhoes:


robhoes:
@johnelse Thanks. I've now changed the first patch to use scanf. Much better :)

johnelse:
Yeah looks good! :shipit:

PR #68, opened 2016-02-22
CA-199405: Set DNS before making DHCP request

simonjbeaumont:
A recent change[1] to support clearing of DNS entries when using Static
mode networking has caused DNS entries to be cleared when using DHCP.

We need to support the following cases:

1. DHCP;
2. Static with non-empty DNS; and
3. Static with empty DNS.

We currently don't support DHCP with a DNS override so we don't need to
consider this in this patch.

This patch changes the order of setting the DNS for the host and calling
dhclient so that the above three use-cases are honoured.

[1]: d49bc77 CA-198824: Allow empty DNS when configuring static IP

Signed-off-by: Si Beaumont <simon.beaumont@citrix.com>


robhoes:
:shipit:

PR #71, opened 2016-03-18
CA-203180: Linux bridge: disable IGMP snooping by default

robhoes:


johnelse:
Looks good :shipit:

PR #74, opened 2016-04-12
CA-206909: Write PID file after finishing startup activities

robhoes:


robhoes:
Depends on xapi-project/xcp-idl#119.

PR #76, opened 2016-05-04
CA-209399: Set mac-table-size to 10000 when creating bridges

jonludlam:
Signed-off-by: Jon Ludlam <jonathan.ludlam@citrix.com>


robhoes:
@jonludlam Let's have config file option for this, so that people can easily change the size of the table if needed (default to 10000).

jonludlam:
Good plan

robhoes:
👍

PR #77, opened 2016-05-09
CA-210158: Only set mac-table-size on real (non-VLAN) bridges

robhoes:

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -806,7 +806,11 @@ module Ovs = struct
 			else
 				[]
 		in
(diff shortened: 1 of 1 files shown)


johnelse:
:shipit:

jonludlam:
Jolly good. Why doesn't it work with VLANs?

robhoes:
@jonludlam Because `ovs-vsctl` pretends that a VLAN bridge is a bridge in the `add-br` call, but it reality the VLAN becomes a port on a "parent" bridge.

jonludlam:
Nice :-(

PR #78, opened 2016-05-11
CA-210762: Linux bridge: Always update bond properties

robhoes:
Bond properties such as the bond mode were not updated if the bond already
existed. This caused XenAPI commands such as Bond.set_mode to have no immediate
effect (changes would be implemented only after re-plugging or rebooting).

Also make sure that bond-manipulation functions in Network_utils are idempotent,
as the high-level function call is supposed to be idempotent.

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -14,6 +14,7 @@
 
 open Listext
 open Xstringext
+open Pervasiveext
 open Fun
 open Network_interface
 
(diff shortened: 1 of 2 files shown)


lib/network_utils.ml
```diff
@@ -214,6 +214,14 @@ info "Found at [ %s ]" (String.concat ", " (List.map string_of_int indices));
 		if is_up dev then
 			ignore (link_set dev ["down"])
 
+	let with_link_down dev f =
+		let up = is_up dev in
+		if up then
+			ignore (link_set dev ["down"]);
+		f ();
+		if up then
+			link_set_up dev
```
johnelse: `link_set_up dev` won't run if `f ()` throws an exception - does that matter?

johnelse:
Could this be two separate commits, since it's essentially making two conceptually different changes?

lib/network_utils.ml
robhoes: I'd better wrap it in `finally` indeed.

robhoes:
I thought about making it two commits, but they would both be necessary for the same fix (making the utils functions idempotent is necessary). That's why I thought that conceptually it just one change.

lib/network_utils.ml
```diff
@@ -402,6 +410,74 @@ module Linux_bonding = struct
...
+		else
+			error "Bond %s does not exist; cannot set slaves" master
+
+	let with_slaves_removed master f =
+		if is_bond_device master then
+			try
+				let slaves = get_bond_slaves master in
+				remove_bond_slaves master slaves;
+				f ();
+				add_bond_slaves master slaves
```
johnelse: Could do with a `finally` here as well.

robhoes:
@johnelse I've added a fixup patch (will squash them before merging).

johnelse:
Ok, I think this looks good now.

PR #79, opened 2016-05-19
CA-211448: When reading a line from sysfs, translate End_of_file to ""

robhoes:
Some sysfs files are "empty", for a good reason.

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -82,7 +82,9 @@ module Sysfs = struct
 			let result = input_line inchan in
 			close_in inchan;
 			result
-		with exn -> close_in inchan; raise (Read_error file)
+		with
+		| End_of_file -> close_in inchan; ""
+		| exn -> error "%s" (Printexc.to_string exn); close_in inchan; raise (Read_error file)
 
 	let write_one_line file l =
 		let outchan = open_out file in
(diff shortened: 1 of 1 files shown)


lib/network_utils.ml
```diff
@@ -82,7 +82,9 @@ module Sysfs = struct
 			let result = input_line inchan in
 			close_in inchan;
 			result
-		with exn -> close_in inchan; raise (Read_error file)
+		with
+		| End_of_file -> close_in inchan; ""
+		| exn -> error "%s" (Printexc.to_string exn); close_in inchan; raise (Read_error file)
```
johnelse: Looks OK, but might be neater as

```
let read_one_line file =
  let inchan = open_in file in
  try
    finally
      (fun () -> input_line inchan)
      (fun () -> close_in inchan)
  with
    | End_of_file -> ""
    | exn -> error "%s" (Printexc.to_string exn); raise (Read_error file)
```

lindig: The `open_in` could also fail. So it could also go inside the `try` and the error could be handled in the `with` block.

robhoes: All agreed; will update.

robhoes:
@johnelse I've added a fixup. Will squash it before merging.

johnelse:
Looks good!

PR #81, opened 2016-05-31
CA-212083: In with_slaves_removed, call `f` if there are no slaves

robhoes:


johnelse:
:shipit:

PR #82, opened 2016-06-20
Don't collect stats for loopback and bond devices

robhoes:
We don't use the stats for these devices, so lets not spend time on them.

Furthermore, due to the fact that the Sysfs.read_one_line and therefore the
Sysfs.get_pci_ids function now log errors, and the loopback and bond devices
do not have PCI info in sysfs, this was causing lots of log spam.


johnelse:
:shipit: :shipit:

PR #85, opened 2016-06-28
CA-211108: Don't consider the number beside the bond mode in sysfs

sharady:
The bond mode node in sysfs includes a number beside the
name of the mode. We should not consider this number while
comparing current-state with desired-state to re-apply the mode.

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -487,7 +487,10 @@ module Linux_bonding = struct
 		if is_bond_device master then begin
 			let get_prop prop =
 				try
-					Some (prop, Sysfs.read_one_line (Sysfs.getpath master ("bonding/" ^ prop)))
+					let bond_prop = Sysfs.read_one_line (Sysfs.getpath master ("bonding/" ^ prop)) in
+					if prop = "mode" then
+						Some (prop, List.hd (String.split ' ' bond_prop))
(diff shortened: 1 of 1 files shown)


lib/network_utils.ml
```diff
@@ -487,7 +487,9 @@ module Linux_bonding = struct
 		if is_bond_device master then begin
 			let get_prop prop =
 				try
-					Some (prop, Sysfs.read_one_line (Sysfs.getpath master ("bonding/" ^ prop)))
+					if prop = "mode" then
+						Some (prop, List.hd (String.split ' ' (Sysfs.read_one_line (Sysfs.getpath master ("bonding/" ^ prop)))))
+					else Some (prop, Sysfs.read_one_line (Sysfs.getpath master ("bonding/" ^ prop)))
```
lindig: I think `(Sysfs.read_one_line (Sysfs.getpath master ("bonding/" ^ prop)))` should be let-bound before the if-expression, which then becomes simpler. Overall this feels as if the format of some configuration file has changed in a way that wasn't anticipated and now the code here has to deal with it in an ad-hoc fashion.

sharady: Thanks @lindig, That was a good suggestion. I have updated the PR.

PR #87, opened 2016-08-18
CP-15132: Systemd conversion

rafalmiel:
Notify systemd when the daemon is ready

Signed-off-by: Rafal Mielniczuk <rafal.mielniczuk@citrix.com>


networkd/networkd.ml
```diff
@@ -108,6 +110,8 @@ let _ =
 		Debug.with_thread_associated "main" start server
 	) ();
 
+        ignore (Daemon.notify Daemon.State.Ready);
```
robhoes: Whitespace...

robhoes:
:shipit: apart from the minor issue. This one definitely needs "notify" for the moment.

robhoes:
Why is the CA-201728 patch included here?

networkd/networkd.ml
```diff
@@ -80,6 +80,8 @@ let doc = String.concat "\n" [
 	"This service looks after host network configuration, including setting up bridges and/or openvswitch instances, configuring IP addresses etc.";
 ]
 
+exception Notify_failed
```
robhoes: Why is this exception defined here? It's not there in the other daemons.

rafalmiel: It is fix to one of the problems that arised after changes. I will remove the exception. It's not needed indeed

PR #88, opened 2016-08-24
CA-219257: Make enic driver workaround configurable.

sharady:
From xcp-networkd.conf file we can configure enic-workaround-until-version="x.x.x.x"
to the version till enic driver workaround will be applied or
set the version to an empty string for not applying the workaround.

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -74,6 +74,13 @@ module Sysfs = struct
 			warn "Failed to obtain list of drivers from sysfs";
 			[]
 
+	let get_driver_version driver () =
+		try
+			Some (String.strip String.isspace (Unixext.string_of_file ("/sys/bus/pci/drivers/" ^ driver ^ "/module/version")))
+		with _ ->
+			warn "Failed to obtain driver version from sysfs";
+			None
+
 	let getpath dev attr =
 		Printf.sprintf "/sys/class/net/%s/%s" dev attr
 
--- a/networkd/network_server.ml
+++ b/networkd/network_server.ml
@@ -27,6 +27,7 @@ type context = unit
 let network_conf = ref "/etc/xcp/network.conf"
 let config : config_t ref = ref empty_config
 let backend_kind = ref Openvswitch
+let enic_workaround_until_version = ref "2.3.0.30"
 
 let legacy_management_interface_start () =
 	try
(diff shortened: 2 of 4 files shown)


lib/network_utils.ml
```diff
@@ -74,6 +74,13 @@ module Sysfs = struct
 			warn "Failed to obtain list of drivers from sysfs";
 			[]
 
+	let get_driver_version driver () =
+		try
+			String.strip String.isspace (Unixext.string_of_file ("/sys/bus/pci/drivers/" ^ driver ^ "/module/version"))
+		with _ ->
+			warn "Failed to obtain driver version from sysfs";
+			""
```
robhoes: Using an option type would be better.

lindig:
I don't have a string opinion about this but would use a different name for the flag: `enic_workaround_enabled=true/false`. It is shorter and shares with the other flag the same `enic` prefix.

robhoes:
The `enic_driver_version` config option should have a name that more clearly indicates what it is for, e.g. `enic-workaround-until-version`. We can also combine the two options into one, and say that the workaround is disabled if `enic-workaround-until-version` is the empty string.

networkd/network_server.ml
```diff
@@ -85,12 +87,22 @@ let set_dns_interface _ dbg ~name =
 	debug "Setting DNS interface to %s" name;
 	config := {!config with dns_interface = Some name}
 
+let compare_driver_version vs1 vs2 () =
+	try
+		let version_of vs = List.map int_of_string (String.split '.' vs) in
+		let first_non_zero lst = List.fold_left (fun a b -> if (a<>0) then a else b) 0 lst in
+		(first_non_zero (List.map2 compare (version_of vs1) (version_of vs2)) <= 0)
```
robhoes: This may not do the right thing if the lists are not the same size.

robhoes: I think this function is a bit hard to follow, and it took me a while to figure out how it is meant to work. Also, it is not obvious what the return value of the function actually means. Also, it looks like `version_of` is really more like a `list_of_version`...

How about something like this:
```
(* Returns `true` is vs1 is older than vs2 *)
let is_older_version vs1 vs2=
  let list_of_version vs = List.map int_of_string (String.split '.' vs) in
  let rec loop vs1' vs2' =
    match vs1', vs2' with
    | [], _ | _, [] -> false
    | a :: _, b :: _ when a < b -> true
    | _ :: tl1, _ :: tl2 -> loop tl1 tl2
  in
  loop (list_of_version vs1) (list_of_version vs2)
```

sharady: This code looks more relaxed for comparing driver version in case version string having different count of digits. Thanks @robhoes I have updated the PR as per review comments.

PR #89, opened 2016-10-07
Set no-flood on PVS-proxy ports

robhoes:
This adds some protection against unicast floods (which happen as part of MAC
learning) getting out of hand.


robhoes:
@anoobs Looks good?

anoobs:
Looks good.

PR #90, opened 2016-10-07
Change monitor blacklist into whitelist

robhoes:
This way we have tighter control over what interfaces are monitored, which are
essentially just PIFs (eth*) and VIFs (vifx.y). Recently, we have had to extend
the blacklist several times, because devices with different names appeared,
which we do not want to monitor.

### DIFF
--- a/networkd/network_monitor_thread.ml
+++ b/networkd/network_monitor_thread.ml
@@ -25,15 +25,9 @@ open D
 (** Table for bonds status. *)
 let bonds_status : (string, (int * int)) Hashtbl.t = Hashtbl.create 10
 
-let monitor_blacklist = ref [
-	"dummy";
-	"xenbr";
-	"xapi";
-	"ovs-system";
-	"xenapi";
-	"lo";
-	"bond";
-	"pvs";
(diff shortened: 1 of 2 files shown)


networkd/network_monitor_thread.ml
```diff
@@ -113,10 +107,11 @@ let get_link_stats () =
 		List.map (fun link ->
 			(standardise_name (Link.get_name link)), link
 		) links |>
+		(* Only keep interfaces with prefixes on the whitelist, and exclude VLAN
+		   devices (ethx.y). *)
 		List.filter (fun (name,link) ->
-			let is_monitor_blacklisted = List.exists (fun s -> String.startswith s name) !monitor_blacklist ||
-						(String.startswith "eth" name && String.contains name '.') in
-			not is_monitor_blacklisted
+			List.exists (fun s -> String.startswith s name) !monitor_whitelist
```
lindig: It is simple enough but I would find creating two predicates even clearer: `is_whitelisted` and `is_vlan` and then use `List.filter (fun (name,_) -> is_whitelisted name && not (is_vlan name)`.

PR #91, opened 2016-10-14
Call mod-port on parent bridge, not "fake" VLAN bridge

robhoes:


robhoes:
@anoobs: please check.

robhoes:
@anoobs says LGTH.

PR #96, opened 2017-01-23
CA-239919: Static IP briefly flushed when a Static IPv4 is set

mseri:
Redefine `Interface.set_ipv4_conf` to keep being idempotent but avoid flushing ip addresses on xapi restart if their ips are the expected ones.


robhoes:
This is not just during a master-slave failover, but each time a static IPv4 address is set. Besides, networkd doesn't know anything about HA and xapi restarts. Please update the commit message to be more specific to networkd.

mseri:
Updated commit title and messages (and PR title)

mseri:
Fixed the issue also for IPv6 as discussed together.

PR #97, opened 2017-02-03
CA-223676: Add function `get_physical_interfaces` to networkd

sharady:
This function will provide the physical interfaces underneath
the bridge.
Bridge can be on:
1) Physical interface.
2) VLAN on physical interface.
3) On a bond.
4) VLAN on a bond.

### DIFF
--- a/networkd/network_server.ml
+++ b/networkd/network_server.ml
@@ -788,11 +788,7 @@ module Bridge = struct
 		match !backend_kind with
 		| Openvswitch ->
 			ignore (Ovs.create_port ~internal:true name bridge);
-			let real_bridge =
-				match Ovs.bridge_to_vlan bridge with
-				| Some (parent, _) -> parent
-				| None -> bridge
-			in
+			let real_bridge = Ovs.get_real_bridge bridge in
 			Ovs.mod_port real_bridge name "no-flood";
 			Interface.bring_up () dbg ~name
 		| Bridge ->
@@ -842,6 +838,31 @@ module Bridge = struct
 				Ovs.bridge_to_interfaces name
 			| Bridge ->
 				Sysfs.bridge_to_interfaces name
+		) ()
+
+	let get_physical_interfaces _ dbg ~name =
+		Debug.with_thread_associated dbg (fun () ->
+			match !backend_kind with
+			| Openvswitch ->
+				Ovs.get_real_bridge name
+				|> Ovs.bridge_to_interfaces
+				|> List.filter (Sysfs.is_physical)
+
+			| Bridge ->
+				let ifaces = Sysfs.bridge_to_interfaces name in
+				let vlan_ifaces = List.filter (fun (bridge, _, _) -> List.mem bridge ifaces) (Proc.get_vlans ()) in
+				let bond_ifaces = List.filter (fun iface -> Linux_bonding.is_bond_device iface) ifaces in
+				let physical_ifaces = List.filter (fun iface -> Sysfs.is_physical iface) ifaces in
+				if vlan_ifaces <> [] then
+					let _, _, parent = List.hd vlan_ifaces in
+					if Linux_bonding.is_bond_device parent then
+						Linux_bonding.get_bond_slaves parent
+					else
+						[parent]
+				else if bond_ifaces <> [] then
+					Linux_bonding.get_bond_slaves (List.hd bond_ifaces)
+				else
+					physical_ifaces
 		) ()
 
 	let get_fail_mode _ dbg ~name =
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -776,6 +776,11 @@ module Ovs = struct
 			debug "bridge_to_vlan: %s" (Printexc.to_string e);
 			None
 
+	let get_real_bridge name =
+		match bridge_to_vlan name with
+		| Some (parent, vlan) -> parent
+		| None -> name
+
 	let get_bond_link_status name =
 		try
 			let raw = appctl ["bond/show"; name] in


networkd/network_server.ml
```diff
@@ -844,6 +844,35 @@ module Bridge = struct
...
+				if ifaces <> [] then ifaces
+				else
+					begin match (Ovs.bridge_to_vlan name) with
+					| Some (parent, vlan) ->
+						let ifaces = Ovs.bridge_to_interfaces parent in
+						List.filter (fun iface -> String.startswith "eth" iface) ifaces
+					| None -> []
+				end
+			| Bridge ->
+				let iface = List.hd (List.filter (fun iface ->
```
lindig: The `List.hd` can fail (similar below). I think it would be better to use pattern matching here to access the first element and to deal with the situation when the list is empty.

networkd/network_server.ml
```diff
@@ -844,6 +844,35 @@ module Bridge = struct
 				Sysfs.bridge_to_interfaces name
 		) ()
 
+	let get_physical_interfaces _ dbg ~name =
+		Debug.with_thread_associated dbg (fun () ->
+			match !backend_kind with
+			| Openvswitch ->
+				let ifaces = List.filter (fun iface ->
```
lindig: This is a suggestion for slightly more compact code:

```
Ovs.bridge_to_interfaces name
|> List.filter (String.startswith "eth")
|> function
| [] ->
  ( match Ovs.bridge_to_vlan name with
    | Some (parent, vlan) ->
      Ovs.bridge_to_interfaces parent
      |> List.filter (String.startswith "eth")
    | None -> []
  )
| ifaces -> ifaces
```

networkd/network_server.ml
```diff
@@ -844,6 +844,35 @@ module Bridge = struct
...
+					String.startswith "eth" iface) (Ovs.bridge_to_interfaces name) in
```
robhoes: Don't use the device name to determine physicalness. This may turn out unreliable. Use `Sysfs.is_physical`.

networkd/network_server.ml
robhoes: You can avoid some duplication by first figuring out the real bridge with
```
let real_bridge =
	match Ovs.bridge_to_vlan bridge with
	| Some (parent, _) -> parent
	| None -> bridge
in
```
and then just once
```
Ovs.bridge_to_interfaces real_bridge
|> List.filter (Sysfs.is_physical)
```
Even better, but a helper function to get the "real" bridge in network_utils (module Ovs), because it is also used elsewhere.

networkd/network_server.ml
robhoes: I agree. And more importantly, don't use the device names to determine the device kind.

networkd/network_server.ml
```diff
@@ -844,6 +844,32 @@ module Bridge = struct
...
+		Debug.with_thread_associated dbg (fun () ->
+			match !backend_kind with
+			| Openvswitch ->
+				Ovs.get_real_bridge name
+				|> Ovs.bridge_to_interfaces
+				|> List.filter (Sysfs.is_physical)
+
+			| Bridge ->
+				let iface =
+					match List.filter (fun iface -> not (String.startswith "vif" iface)) (Sysfs.bridge_to_interfaces name) with
```
sharady: Samples of bridges on bridge backend:
[root@xrtmia-06-07 brif]# ls /sys/class/net/xenbr2/brif/
eth2  vif1.0
[root@xrtmia-06-07 brif]# ls /sys/class/net/xapi1/brif/
bond0  vif1.1
[root@xrtmia-06-07 net]# ls /sys/class/net/xapi2/brif
bond0.3201  vif1.2
[root@xrtmia-06-07 net]# ls /sys/class/net/xapi0/brif
eth2.3201

I am not able to found any existing function which can collect vlan interfaces from bridge.
(Used `vif` to collect them).

sharady: Thanks @robhoes @lindig for the review, Please have a look.

robhoes: You'll need a different way to determine whether an interface is a VLAN interface in the Linux kernel. This is separate from the bridge stuff (unlike for the OVS).

networkd/network_server.ml
```diff
@@ -844,6 +844,32 @@ module Bridge = struct
 				Sysfs.bridge_to_interfaces name
 		) ()
 
+	let get_physical_interfaces _ dbg ~name =
+		Debug.with_thread_associated dbg (fun () ->
+			match !backend_kind with
+			| Openvswitch ->
+				Ovs.get_real_bridge name
```
robhoes: Can you change the other function above to use the new `get_real_bridge` as well please.

networkd/network_server.ml
robhoes: In fact, you're already using `Proc.get_vlans` below to find VLAN interfaces.

sharady:
Cool!! Got a way to identify the interface type without using `vif` also.
Thanks @robhoes, please have a look now.

networkd/network_server.ml
```diff
@@ -839,11 +839,37 @@ module Bridge = struct
 		Debug.with_thread_associated dbg (fun () ->
 			match !backend_kind with
 			| Openvswitch ->
-				Ovs.bridge_to_interfaces name
+				Ovs.get_real_bridge name
+				|> Ovs.bridge_to_interfaces
```
robhoes: I didn't mean to change this function. This is wrong, because it changes the semantics. Instead, the function `add_pvs_proxy_port` has some lines that can be replaced by the new function.

robhoes:
This is good now, apart from the one thing above.

networkd/network_server.ml
sharady: Sorry!!

PR #100, opened 2017-05-12
CP-22034: read PIF speed & duplex from sysfs

thomassa:


thomassa:
I wondered about replacing `get_status` with a pair of functions `get_speed` and `get_duplex`, but I decided against it: it would not simplify the code in `network_monitor_thread.ml` that currently uses the `get_status` function.

_oasis
```diff
@@ -28,7 +28,6 @@ Library networklibs
     xcp-inventory,
     xcp.network,
     systemd
-  CSources:           link_stubs.c, netdev.h
   CCOpt:              -Wno-unused-function -g -ggdb
```
thomassa: The `CCOpt` line is harmless, but I'm not sure whether or not it would be better to remove it.

robhoes: If all C bindings are gone, then let's remove this line as well.

thomassa:
I've removed the `CCOpt` line now.

PR #101, opened 2017-05-18
CA-254379: Fixing vlan bridge create during host installation.

sharady:
Host installer doesn't write management_interface into inventory.
As firstboot needs to create management pif and update the inventory.

In case of management on vlan xcp-networkd must not create vlan bridge
if management_interface is empty.
Let the firstboot create the VLAN and set the management PIF.

### DIFF
--- a/lib/network_config.ml
+++ b/lib/network_config.ml
@@ -76,20 +76,24 @@ let read_management_conf () =
 				ports = [device, {default_port with interfaces = [device]}];
 				persistent_b = true
 				} in
-			match vlan with
-			| None ->
-				[device, phy_interface; bridge_name, bridge_interface],
-				[bridge_name, primary_bridge_conf]
-			| Some vlan ->
-				let parent = bridge_naming_convention device in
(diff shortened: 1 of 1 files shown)


lib/network_config.ml
```diff
@@ -76,20 +76,25 @@ let read_management_conf () =
 				ports = [device, {default_port with interfaces = [device]}];
 				persistent_b = true
 				} in
-			match vlan with
-			| None ->
+			if bridge_name = "" then
 				[device, phy_interface; bridge_name, bridge_interface],
```
robhoes: These two lines are the same as the lines below in the `None` case. That could be deduplicated.
But more importantly, `bridge_name` is `""` here, so why is it used at all? Shouldn't we just leave the entire config and return `[], []`?

sharady: make sense, I tried to retain the older behaviour but not needed though.

robhoes: Well, I don't think it would work, because you can have interfaces/bridges called "".

PR #102, opened 2017-05-19
CP-22405: Log message for config file not existing

Frezzle:
Removes spurious backtrace occurring on first boot
because config file /var/lib/xcp/networkd.db does
not exist yet.

Signed-off-by: Frederico Mazzone <frederico.mazzone@citrix.com>

### DIFF
--- a/lib/network_config.ml
+++ b/lib/network_config.ml
@@ -90,8 +90,11 @@ let read_config () =
 	try
 		let config_json = Unixext.string_of_file config_file_path in
 		config_json |> Jsonrpc.of_string |> config_t_of_rpc
-	with e ->
-		error "Error while trying to read networkd configuration: %s\n%s"
-			(Printexc.to_string e) (Printexc.get_backtrace ());
-		raise Read_error
-
+	with
+		| Unix.Unix_error (Unix.ENOENT, _, file) ->
+			info "Cannot read networkd configuration file %s because it does not exist." file;
+			raise Read_error
+		| e ->
+			info "Error while trying to read networkd configuration: %s\n%s"
+				(Printexc.to_string e) (Printexc.get_backtrace ());
+			raise Read_error


Frezzle:
Before the change, the log would show this on first-boot:
```
May 16 08:42:21 localhost xcp-networkd: [error|localhost|0 |startup|network_config] Error while trying to read networkd configuration: Unix.Unix_error(Unix.ENOENT, "open", "/var/lib/xcp/networkd.db")#012Raised by primitive operation at file "lib/unixext.ml", line 120, characters 10-39#012Called from file "lib/unixext.ml", line 170, characters 47-73#012Called from file "lib/network_config.ml", line 91, characters 20-59
```
After the change, we do not have a stacktrace for the case of the file not existing:
```
May 18 18:57:27 localhost xcp-networkd: [error|dt53|0 |startup|network_config] Cannot read networkd configuration file /var/lib/xcp/networkd.db because it does not exist.
```

I tested this by temporarily removing the file before reading it, to simulate the first-boot scenario.

lib/network_config.ml
```diff
@@ -90,8 +90,12 @@ let read_config () =
...
 		config_json |> Jsonrpc.of_string |> config_t_of_rpc
-	with e ->
-		error "Error while trying to read networkd configuration: %s\n%s"
-			(Printexc.to_string e) (Printexc.get_backtrace ());
-		raise Read_error
+	with
+		| Unix.Unix_error (Unix.ENOENT, _, file) ->
+			error "Cannot read networkd configuration file %s because it does not exist." file;
+			raise Read_error
+		| _ as e ->
```
gaborigloi: This can be written as simply `| e ->`, if I'm not mistaken, or does using `_` in this case have any significance?

Frezzle: Yes, it's the same when I tested. But I was wondering what would happen if some variable or type in the future were created named `e` and then it matched here accidentally. I am being overly-explicit out of caution, or is there no need as the case I've described can't happen?

gaborigloi: Oh, I didn't think of that. I think we can't declare lowercase exceptions though.

gaborigloi: But it looks nice and explicit, so I'm happy if we leave it as it is :) (and sorry for nitpicking)

lib/network_config.ml
```diff
@@ -90,8 +90,12 @@ let read_config () =
 	try
 		let config_json = Unixext.string_of_file config_file_path in
 		config_json |> Jsonrpc.of_string |> config_t_of_rpc
-	with e ->
-		error "Error while trying to read networkd configuration: %s\n%s"
-			(Printexc.to_string e) (Printexc.get_backtrace ());
-		raise Read_error
+	with
+		| Unix.Unix_error (Unix.ENOENT, _, file) ->
+			error "Cannot read networkd configuration file %s because it does not exist." file;
```
mseri: Is it actually an error or we re-create the configuration file when it is not present? I think this should be a warning, it seems to me that we log the `Read_error` as an error on the various call site.

lib/network_config.ml
mseri: The `try/with` will shadow any pre-defined `e` here.
If you had defined an `e` and want to pattern match on it you would have to do `e when e = e` (and somebody would make you change it before the PR is accepted ;))

lib/network_config.ml
robhoes: Indeed, it should not be an error. When starting, the daemon tries to read and parse that file, but it may not be there (e.g. on first boot). Make it `info`.

mseri: I think it's worth looking at the places where the function is called and check the logging used there.

lib/network_config.ml
robhoes: Well, I do agree with Gabor :) I would prefer `| e ->`, because it is shorter, and exceptions always start with an upper-case letter, so there won't be any conflicts.

lib/network_config.ml
Frezzle: Are there conditions/guidelines as to which things you do/don't make into either error, info or debug messages?

robhoes: https://en.wikipedia.org/wiki/Syslog#Severity_level

PR #107, opened 2017-07-07
CP-23093: Implement toggle of IGMP snooping on xcp-networkd

fillzero:
Signed-off-by: Liang Dai <liang.dai1@citrix.com>

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -867,7 +867,7 @@ module Ovs = struct
 			with _ -> ());
 		) phy_interfaces
 
-	let create_bridge ?mac ?external_id ?disable_in_band ~fail_mode vlan vlan_bug_workaround name =
+	let create_bridge ?mac ?external_id ?disable_in_band ?igmp_snooping ~fail_mode vlan vlan_bug_workaround name =
 		let vlan_arg = match vlan with
 			| None -> []
 			| Some (parent, tag) ->
@@ -917,14 +917,12 @@ module Ovs = struct
 			else
 				[]
 		in
-		let set_mcast_snooping =
-			if vlan = None then
-				["--"; "set"; "bridge"; name; "mcast_snooping_enable=true"]
-			else
-				[]
+		let set_igmp_snooping = match igmp_snooping, vlan with
+			| Some x, None -> ["--"; "set"; "bridge"; name; "mcast_snooping_enable=" ^ (string_of_bool x)]
+			| _ -> []
 		in
 		vsctl ~log:true (del_old_arg @ ["--"; "--may-exist"; "add-br"; name] @
-			vlan_arg @ mac_arg @ fail_mode_arg @ disable_in_band_arg @ external_id_arg @ vif_arg @ set_mac_table_size @ set_mcast_snooping)
+			vlan_arg @ mac_arg @ fail_mode_arg @ disable_in_band_arg @ external_id_arg @ vif_arg @ set_mac_table_size @ set_igmp_snooping)
 
 	let destroy_bridge name =
 		vsctl ~log:true ["--"; "--if-exists"; "del-br"; name]
--- a/networkd/network_server.ml
+++ b/networkd/network_server.ml
@@ -546,14 +546,14 @@ module Bridge = struct
 			) existing_bridges
 		end
 
-	let create _ dbg ?vlan ?mac ?(other_config=[]) ~name () =
+	let create _ dbg ?vlan ?mac ?igmp_snooping ?(other_config=[]) ~name () =
 		Debug.with_thread_associated dbg (fun () ->
 			debug "Creating bridge %s%s" name (match vlan with
 				| None -> ""
 				| Some (parent, vlan) -> Printf.sprintf " (VLAN %d on bridge %s)" vlan parent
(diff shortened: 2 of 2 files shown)


fillzero:
@robhoes -- Would you please review this PR?
Based on the multicast feature design, we need to implement a toggle of IGMP Snooping.

fillzero:
This PR depends on https://github.com/xapi-project/xcp-idl/pull/164

networkd/network_server.ml
```diff
@@ -983,10 +989,10 @@ module Bridge = struct
 			debug "** Configuring the following bridges: %s%s"
 				(String.concat ", " (List.map (fun (name, _) -> name) config))
 				(if conservative then " (best effort)" else "");
-			List.iter (function (bridge_name, ({ports; vlan; bridge_mac; other_config; _} as c)) ->
+			List.iter (function (bridge_name, ({ports; vlan; bridge_mac; igmp_snooping; other_config; _} as c)) ->
 				update_config bridge_name c;
 				exec (fun () ->
-					create () dbg ?vlan ?mac:bridge_mac ~other_config ~name:bridge_name ();
+					create () dbg ?vlan ?mac:bridge_mac ?igmp_snooping:igmp_snooping ~other_config ~name:bridge_name ();
```
krizex: use `~igmp_snooping` better than `?igmp_snooping:igmp_snooping`

networkd/network_server.ml
```diff
@@ -949,6 +949,12 @@ module Bridge = struct
 			update_config name {(get_config name) with persistent_b = value}
 		) ()
 
+	let set_ovs_igmp_snooping _ dbg ~name ~value =
```
robhoes: We don't need this function, because we can just call `create` even if the bridge already exists. All functions in the interface, including `make_config` and `create` are idempotent, which means that you can safely call them multiple times. The job of `Bridge.make_config` is to update the current configuration of the bridge based on the new configuration that is given, and only do something when (part of) the configuration has changed.

networkd/network_server.ml
robhoes: `?igmp_snooping` probably.

networkd/network_server.ml
fillzero: @robhoes Thanks for your comments, updated PR accordingly

networkd/network_server.ml
```diff
@@ -546,14 +546,14 @@ module Bridge = struct
...
-	let create _ dbg ?vlan ?mac ?(other_config=[]) ~name () =
+	let create _ dbg ?vlan ?mac ?igmp_snooping ?(other_config=[]) ~name () =
 		Debug.with_thread_associated dbg (fun () ->
 			debug "Creating bridge %s%s" name (match vlan with
 				| None -> ""
 				| Some (parent, vlan) -> Printf.sprintf " (VLAN %d on bridge %s)" vlan parent
 			);
 			Stdext.Opt.iter (destroy_existing_vlan_bridge name) vlan;
-			update_config name {(get_config name) with vlan; bridge_mac=mac; other_config};
+			update_config name {(get_config name) with vlan; bridge_mac=mac; igmp_snooping=igmp_snooping; other_config};
```
robhoes: This can be just `igmp_snooping` (because the parameter has the same name).

lib/network_utils.ml
```diff
@@ -917,14 +917,16 @@ module Ovs = struct
...
-			if vlan = None then
-				["--"; "set"; "bridge"; name; "mcast_snooping_enable=true"]
-			else
+		let set_igmp_snooping = match igmp_snooping with
+			| None -> []
+			| Some igmp_snooping ->
+				if vlan = None then
+					["--"; "set"; "bridge"; name; "mcast_snooping_enable=" ^ (string_of_bool igmp_snooping)]
+				else
 				[]
```
robhoes: One more tab needed.

robhoes: In fact, we can easily combine the match and if:
```
match igmp_snooping, vlan with
| Some x, None -> ["--"; "set"; "bridge"; name; "mcast_snooping_enable=" ^ (string_of_bool x)]
| _ -> []
```

fillzero: Thanks. Updated accordingly.

robhoes:
This is fine now, but we can't merge it without the matching change in xcp-idl, and we can't merge _that_ without a change to xen-api, for which there is no PR yet.

fillzero:
Thanks @robhoes . Regarding the xen-api change, @YarsinCitrix will send PR later. 

Also copy @huizh to be aware.

robhoes:
@fillzero @huizh  For the xen-api change, I would suggest to first have a PR to _just_ fill in the new field in the network_config with a default value, so that these changes can go in quickly. We can have a separate PR for the API changes, for which we are still discussing the design.

robhoes:
I'll close this for now. When the xen-api patch is ready, I'll reopen and merge this PR.

huizh:
Hi @robhoes, thanks for the review and the suggestion. Very helpful. Will follow up your suggestion to raise a XenAPI PR so that to get this and https://github.com/xapi-project/xcp-idl/pull/164 merged.

PR #109, opened 2017-08-17
CP-23843: Disable IPv6 multicast snooping for OVS in XenServer

krizex:
CP-23098 implemented a new OVS toggle `enable-ipv6-mcast-snooping` and set its default value to `true` for ensuring compatible with OVS.
This PR make the default value of this toggle to `false` because XenServer do not support IPv6 multicast snooping.


krizex:
@robhoes Please help review, thank you!

PR #111, opened 2017-08-24
Implement UT for ovs new configuration parameters

krizex:
PR includes:
CP-24075 Implement UT for disable IPv6 igmp snooping
CP-24076 Implement UT for disable flood of unregistered traffic

### DIFF
--- a/test/network_test_ovs.ml
+++ b/test/network_test_ovs.ml
@@ -1,0 +1,84 @@
+(*
+ * Copyright (C) 2017 Citrix Systems Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


test/network_test_ovs.ml
```diff
@@ -0,0 +1,84 @@
+(*
+ * Copyright (C) 2006-2017 Citrix Systems Inc.
```
lindig: How can Citrix claim copyright from 2006 onwards for code that didn't exist at the time? Just use 2017.

krizex: Thanks for your reminder, this section is copied from other files, unfortunately I omit that.

robhoes:
As discussed, I don't think that this is worth doing.

PR #112, opened 2017-08-29
CP-24074 Implement UT for inject igmp query when turn on igmp snooping

YarsinCitrix:
Added three Unit test cases to test OVS.create_bridge, OVS.get_mcast_snooping_enable and OVS.inject_igmp_query.

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -39,7 +39,15 @@ let inject_igmp_query_script = ref "/usr/libexec/xenopsd/igmp_query_injector.py"
 let mac_table_size = ref 10000
 let igmp_query_maxresp_time = ref "5000"
 
+let in_unit_test = ref false
+let unit_test_call_script = ref ""
+
 let call_script ?(log_successful_output=false) ?(timeout=Some 60.0) script args =
+	if !in_unit_test then begin
+		unit_test_call_script := (String.concat " " (script::args));
+		!unit_test_call_script
+	end
+	else
 	try
 		Unix.access script [ Unix.X_OK ];
 		(* Use the same $PATH as xapi *)
--- a/test/network_test_igmp.ml
+++ b/test/network_test_igmp.ml
@@ -1,0 +1,65 @@
+(*
+ * Copyright (C) 2006-2013 Citrix Systems Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify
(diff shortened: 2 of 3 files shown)


test/network_test_igmp.ml
```diff
@@ -0,0 +1,65 @@
...
+			!OVS_Cli_test.vsctl_output)
+
+(* Test OVS.get_mcast_snooping_enable to form the correct parameter to query OVS *)
+let test_get_mcast_snooping_enable () = 
+	let module Ovs = Ovs.Make(OVS_Cli_test) in
+	let bridge = "xapi1" in
+	let expect = "-- get bridge xapi1 mcast_snooping_enable" in
+	ignore (Ovs.get_mcast_snooping_enable bridge);
+
+	assert_bool "Incorrect input format!"
```
lindig: I believe `aasert_equal` would be more idiomatic here. Also below.

test/network_test_igmp.ml
```diff
@@ -0,0 +1,65 @@
+(*
+ * Copyright (C) 2006-2013 Citrix Systems Inc.
```
lindig: If this is a new file, why is it Copyright 2006-2013?

lib/network_utils.ml
```diff
@@ -39,7 +39,15 @@ let inject_igmp_query_script = ref "/usr/libexec/xenopsd/igmp_query_injector.py"
 let mac_table_size = ref 10000
 let igmp_query_maxresp_time = ref "5000"
 
+let in_unit_test = ref false
```
lindig: Is this reference manipulated from outside? Where? If not, this does not have to be a reference.

lib/network_utils.ml
```diff
@@ -39,7 +39,15 @@ let inject_igmp_query_script = ref "/usr/libexec/xenopsd/igmp_query_injector.py"
...
+let unit_test_call_script = ref ""
+
 let call_script ?(log_successful_output=false) ?(timeout=Some 60.0) script args =
+	if !in_unit_test then begin
```
robhoes: This is all a bit odd. If we really want to do this, then we should functorise this bit as done for the Ovs module. But as I said in my other comment, I don't think this is really worth it at all.

robhoes:
Thanks for the effort, but I'm sorry, I am going to close this and suggest to not have unit tests for this.

PR #113, opened 2017-09-11
CA-264980: /usr/libexec/xenopsd/igmp_query_injector.py was broken

YarsinCitrix:

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -41,14 +41,19 @@ let igmp_query_maxresp_time = ref "5000"
 let enable_ipv6_mcast_snooping = ref false
 let mcast_snooping_disable_flood_unregistered = ref true
 
-let call_script ?(log_successful_output=false) ?(timeout=Some 60.0) script args =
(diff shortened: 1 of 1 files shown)


lib/network_utils.ml
```diff
@@ -41,14 +41,19 @@ let igmp_query_maxresp_time = ref "5000"
 let enable_ipv6_mcast_snooping = ref false
 let mcast_snooping_disable_flood_unregistered = ref true
 
-let call_script ?(log_successful_output=false) ?(timeout=Some 60.0) script args =
+let call_script ?(log_successful_output=false) ?(timeout=Some 60.0) ?(detach_child=false) script args =
```
euanh: I don't like the way that this function returns an empty string in the `detach_child` case.  It might be better to have two functions - `call_script`, which blocks and returns the output, and `fork_script` (or something similar) which detaches and returns unit.    These functions would share a lot of logic which could be factored out.   What do you think @robhoes ?

euanh:
Apart from the question to @robhoes, this looks ok to me.

lib/network_utils.ml
robhoes: I agree.

PR #115, opened 2017-10-04
CA-196520: use `mtu_request` in OVS to set interface MTU

edwintorok:
If `ip link set` is used OVS will keep changing it back to the minimum MTU if you attach 2 VLANs (created on a bond) with different MTUs to the same VM.

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -1119,6 +1119,9 @@ module Ovs = struct
 	let mod_port bridge port action =
 		ofctl ~log:true ["mod-port"; bridge; port; action] |> ignore
 
+	let set_mtu interface mtu =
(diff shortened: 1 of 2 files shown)


lib/network_utils.ml
```diff
@@ -1119,6 +1119,9 @@ module Ovs = struct
 	let mod_port bridge port action =
 		ofctl ~log:true ["mod-port"; bridge; port; action] |> ignore
 
+	let set_mtu bridge mtu =
```
robhoes: Better use `interface` rather than `bridge`, because that's what it really is.

lib/network_utils.ml
```diff
@@ -1119,6 +1119,9 @@ module Ovs = struct
...
+	  vsctl ~log:true ["set"; "interface"; bridge; Printf.sprintf "mtu_request=%d" mtu]
```
robhoes: Inconsistent indentation.

PR #116, opened 2017-10-12
CA-268679 XAPI loop report (Sys_error "Invalid argument")

krizex:
Match Sys_error("Invalid argument") and omit the error log.
For other error case, record the exception info and backtrace.

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -109,7 +109,12 @@ module Sysfs = struct
 				(fun () -> close_in inchan)
 		with
 		| End_of_file -> ""
-		| exn -> error "%s" (Printexc.to_string exn); raise (Read_error file)
+		(* Match the exception when the device state if off *)
+		| Sys_error("Invalid argument") -> raise (Read_error file)
+		| exn ->
+			error "Error in read one line of file: %s, exception %s\n%s"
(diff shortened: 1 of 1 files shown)


lib/network_utils.ml
```diff
@@ -109,7 +109,12 @@ module Sysfs = struct
 				(fun () -> close_in inchan)
 		with
 		| End_of_file -> ""
-		| exn -> error "%s" (Printexc.to_string exn); raise (Read_error file)
+		(* Match the exception when the device state if off *)
+		| Sys_error("Invalid argument") -> raise (Read_error file)
```
mseri: This is introducing a new warning:
```
Warning 52: Code should not depend on the actual values of
this constructor's arguments. They are only for information
and may change in future versions. (See manual section 8.5)
```
but I don't think that there is a nicer way of dealing with it at the moment.
The workaround would be using directly unix file descriptors and would make the code less maintainable.

krizex: Can we add some options into Makefile to restrain this warning?

krizex:
Updated

lib/network_utils.ml
mseri: I do not want to restrain this warning. It reminds us that we are using a potentially unstable error match. It's good to have it there

PR #118, opened 2017-12-07
CP-26156: Port xcp-networkd to use jbuilder

YarsinCitrix:


.gitignore
```diff
@@ -6,5 +6,7 @@ xcp-networkd.1
 *.native
 setup.data
 setup.log
+*.merlin
+*.install
 dist/
 _build/
```
mseri: This can be simply
```
_build/
.merlin
*.install
```

test/jbuild
```diff
@@ -0,0 +1,13 @@
+(executable
+ ((name network_test)
+  (public_name network-test)
```
mseri: Not needed for tests, you don't want to install them

Makefile
```diff
@@ -1,67 +1,36 @@
...
-
-coverage: _tags _tags.coverage 
-	test ! -f _tags.orig && mv _tags _tags.orig || true
-	cat _tags.coverage _tags.orig > _tags
+	jbuilder clean
 
-uncover: _tags.orig
-	mv _tags.orig _tags
+test:
+	_build/default/test/network_test.exe
```
mseri: Use 
```
jbuilder runtest
```
and add teh following to `test/jbuild`
```
(alias
 ((name    runtest)
  (deps    (network_test.exe))
  (action  (run ${<}))))
```

Makefile
```diff
@@ -1,67 +1,36 @@
...
-networkd/version.ml: VERSION
-	echo "let version = \"$(shell cat VERSION)\"" > networkd/version.ml
-
-doc: setup.data setup.bin
-	@./setup.bin -doc -j $(J)
-
-test: setup.bin build
-	@./setup.bin -test
+release:
+	jbuilder build @install
```
mseri: Change this to 
```
jbuilder build @install @networkd/man
```
and make this the default make target.

The `--dev` build could fail if there are warnings, we should not silence them and fix the one that can be fixed, but then build the product without the `--dev` flag

YarsinCitrix: I will update for most comments. However, for the library split up thing, I would like to do it separately, tracked by another ticket. In that case, the jbuild porting stuff will not need to wait.

opam
```diff
@@ -1,17 +0,0 @@
-opam-version: "1"
```
mseri: The `opam` file should _not _ be delete, it should be renamed `xapi-networkd.opam`.
You need to delete the `remove: ...` section
You need to replace the `build: ...` section with
```
build: ["jbuilder" "build" "-p" name "-j" jobs]
build-test: ["jbuilder" "runtest" "-p" name]
```

You also need to remove `bisect_ppx` and `ocamlfind` from the `depends:` and add `"jbuilder" {build & >= "1.0+beta11"}` and `"xapi-stdext" {>= 3.0.0}"

cli/jbuild
```diff
@@ -0,0 +1,14 @@
+(executable
+ ((name network_cli)
+  (public_name networkd-cli)
```
mseri: Travis failed:

```
File "cli/jbuild", line 4, characters 2-24:

Error: You cannot declare items to be installed without adding a <package>.opam file at the root of your project.

To declare elements to be installed as part of package "xcp-networkd", add a "xcp-networkd.opam" file at the root of your project.
```

You do need the opam file, I've left a comment there

PR #123, opened 2018-01-10
CP-26216: implement to enable and disable SR-IOV

xiewei20082008:
SRIOV operations
* enable SR-IOV
* disable SR-IOV
* get SR-IOV capability

configurable items:

* MAC address
* VLAN
* rate limiting

Signed-off-by: Wei Xie <wei.xie@citrix.com>

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -13,6 +13,8 @@
  *)
 
 open Xapi_stdext_pervasives
+open Xapi_stdext_unix
+open Xapi_stdext_std
 
 open Network_interface
 
@@ -31,6 +33,11 @@ let brctl = ref "/sbin/brctl"
 let modprobe = "/sbin/modprobe"
 let ethtool = ref "/sbin/ethtool"
 let bonding_dir = "/proc/net/bonding/"
+let lspci = "/sbin/lspci"
+let ip = "/sbin/ip"
+let uname = "/usr/bin/uname"
+let dracut = "/sbin/dracut"
+let rebuild_initrd_timeout = ref 120.0
 let fcoedriver = ref "/opt/xensource/libexec/fcoe_driver"
 let inject_igmp_query_script = ref "/usr/libexec/xenopsd/igmp_query_injector.py"
 let mac_table_size = ref 10000
@@ -208,6 +215,36 @@ module Sysfs = struct
 		|> (fun p -> try read_one_line p |> duplex_of_string with _ -> Duplex_unknown)
 		in (speed, duplex)
 
+	let get_dev_nums_with_same_driver driver = 
+		try
+			Sys.readdir ("/sys/bus/pci/drivers/" ^ driver)
+			|> Array.to_list
+			|> List.filter (Re.execp (Re_perl.compile_pat "\d+:\d+:\d+\.\d+"))
+			|> List.length
+		with _ -> 0
+
+	let parent_device_of_vf pcibuspath =
+		try
+			let pf_net_path = Printf.sprintf "/sys/bus/pci/devices/%s/physfn/net" pcibuspath in
+			let devices = Sys.readdir pf_net_path in
+			if Array.length devices = 1 then devices.(0)
+			else raise (Sys_error ("Can not get parent device for " ^ pcibuspath))
+		with _ -> raise (Sys_error ("Can not get parent device for " ^ pcibuspath))
+
+	let device_index_of_vf parent_device pcibuspath =
+		try
+			let re = Re_perl.compile_pat "virtfn(\d+)" in
+			let device_path = getpath parent_device "device" in
+			let group = Sys.readdir device_path
+				|> Array.to_list
+				|> List.filter (Re.execp re) (* List elements are like "virtfn1" *)
+				|> List.find (fun x -> Xstringext.String.has_substr (Unix.readlink (device_path ^ "/" ^ x)) pcibuspath )
+				|> Re.exec_opt re
+			in
+			match group with
+			| None -> raise (Sys_error ("Can not get device index for " ^ pcibuspath))
+			| Some x -> int_of_string (Re.Group.get x 1)
+		with _ -> raise (Sys_error ("Can not get device index for " ^ pcibuspath))
 end
 
 module Ip = struct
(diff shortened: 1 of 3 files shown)


cheng-z:
Hi @robhoes , I saw a Travis failure which was caused by: Sysfs_successful newly defined in
https://github.com/xapi-project/xcp-idl/pull/185/files
was not included in Travis.

So I think basically, it was not a problem.
If you can let me know how to point the travis for the feature branch of IDL, I can do that to pass the travis check. Thanks.

robhoes:
I am updating the travis config.

lib/network_utils.ml
```diff
@@ -208,6 +215,36 @@ module Sysfs = struct
...
+			|> List.length
+		with _ -> 0
+
+	let parent_device_of_vf pcibuspath =
+		try
+			let pf_net_path = Printf.sprintf "/sys/bus/pci/devices/%s/physfn/net" pcibuspath in
+			let devices = Sys.readdir pf_net_path in
+			if Array.length devices = 1 then devices.(0)
+			else raise (Sys_error ("Can not get parent device for " ^ pcibuspath))
+		with _ -> raise (Sys_error ("Can not get parent device for " ^ pcibuspath))
```
robhoes: The `Sys_error` exception is raised by functions in the `Sys` module. I don't think that is appropriate here. Also, this exception will bubble up all the way to the top. I'd prefer to return an option type from functions like this, so that the control flow is easier to understand. Or even better, use the `result` type and monadic operators from `Rresult`.

Note that generally speaking, the error handling in networkd is quite poor. But it would be good to start right now to make it better in the new code.

lib/network_utils.ml
```diff
@@ -208,6 +215,36 @@ module Sysfs = struct
...
+			let group = Sys.readdir device_path
+				|> Array.to_list
+				|> List.filter (Re.execp re) (* List elements are like "virtfn1" *)
+				|> List.find (fun x -> Xstringext.String.has_substr (Unix.readlink (device_path ^ "/" ^ x)) pcibuspath )
+				|> Re.exec_opt re
+			in
+			match group with
+			| None -> raise (Sys_error ("Can not get device index for " ^ pcibuspath))
+			| Some x -> int_of_string (Re.Group.get x 1)
+		with _ -> raise (Sys_error ("Can not get device index for " ^ pcibuspath))
```
robhoes: As above, regarding the exception.

lib/network_utils.ml
```diff
@@ -1163,3 +1200,196 @@ module Ethtool = struct
...
+				| false, false -> set_sriov_via_sysfs dev 0; ()
+				| true, true ->
+					write_conf_file driver conf;
+					rebuild_initrd ()
+				| _ -> ()
+			with e -> debug "%s: best effort disable sriov with exn: %s" dev (Printexc.to_string e);()
+		in
+		best_effort_disable(); Ok Disable_successful
+
+	let make_vf_conf_internal pcibuspath (vf_info : Sriov.sriov_pci_t) =
```
robhoes: Any code manipulating types from the IDL, such as `Sriov.sriov_pci_t`, should go into network_server.ml. Functions in network_util.ml are meant to be simple wrappers of Linux tools, which are tied together in network_server.ml functions. Hence, this entire function should be part of `make_vf_config`.

lib/network_utils.ml
```diff
@@ -1163,3 +1200,196 @@ module Ethtool = struct
...
+		let config_mac dev index  = function
```
robhoes: These three functions belong in the `Ip` module and call `link_set`.

lib/network_utils.ml
```diff
@@ -1163,3 +1200,196 @@ module Ethtool = struct
 		if options <> [] then
 			ignore (call ~log:true ("-K" :: name :: (List.concat (List.map (fun (k, v) -> [k; v]) options))))
 end
+
+module Sriov = struct
```
robhoes: Note that Network_utils (this file) is organised into sub-modules that are based on Linux tools/features (see comment below on `make_vf_conf_internal`). `Sriov` doesn't quite fit this model. It looks like most of the functions below belong inside some of the existing modules above, or in new ones, or should just move to the top-level in network_server.ml.

lib/network_utils.ml
```diff
@@ -1163,3 +1200,196 @@ module Ethtool = struct
...
+	type t = {
+		max_vfs: int;
+		num_vfs: int;
+	}
+
+	let rebuild_initrd () =
+		let img_name = call_script uname ["-r"] |> String.trim in
+		ignore(call_script ~timeout:(Some !rebuild_initrd_timeout) dracut
```
robhoes: This uses `dracut`, so could go into a `Dracut` module.

lib/network_utils.ml
```diff
@@ -1163,3 +1200,196 @@ module Ethtool = struct
...
+			match group with
+			| None -> {max_vfs=0; num_vfs=0}
+			| Some x -> {max_vfs = (int_of_string (Re.Group.get x 1)) - 1;
+				num_vfs = (int_of_string (Re.Group.get x 2))}
+		end
+		else
+			{max_vfs=0; num_vfs=0}
+
+	let get_lspci_output dev = 
+		call_script lspci ["-vv"; "-s"; (Sysfs.get_pcibuspath dev)]
```
robhoes: We should try to use `ocaml-pci` rather than calling `lspci` and parsing its output.

lib/network_utils.ml
```diff
@@ -1163,3 +1200,196 @@ module Ethtool = struct
...
+			| true, true ->
+				write_conf_file driver conf;
+				rebuild_initrd ()
+			| false, false -> 
+				let new_option_line = Printf.sprintf "options %s max_vfs=%s" driver (gen_options_for_maxvfs driver config_t.max_vfs) in
+				write_conf_file driver (conf @ [new_option_line]);
+				rebuild_initrd ()
+			| _ -> ()
+		in
+		if config_t.max_vfs = 0 then raise (Sys_error (Printf.sprintf "%s: do not have sriov capabilities" dev));
```
robhoes: This function actually returns a result type in the IDL, while this exception is never caught. This should translate to an error that is defined in the IDL.

lib/network_utils.ml
```diff
@@ -1163,3 +1200,196 @@ module Ethtool = struct
...
+
+	let get_driver dev =
+		match (Sysfs.get_driver_name dev) with
+		| Some driver -> driver
+		| None -> raise (Sys_error ("cannot get driver name for " ^ dev))
+
+	let write_conf_file driver content=
+		Unixext.write_string_to_file (Printf.sprintf "/etc/modprobe.d/%s.conf" driver) (String.concat "\n" content)
+
+	let enable_internal dev =
```
robhoes: See the comment on `make_vf_conf_internal` below. Most of this function should be moved into the `enable` function in network_server.ml.

lib/network_utils.ml
```diff
@@ -1163,3 +1200,196 @@ module Ethtool = struct
...
+					Error (Unknown "unable to enable via sysfs")
+			end
+		with
+		| Sys_error s when Xstringext.String.has_substr s "out of range of" -> Error Bus_out_of_range
+		| Sys_error s when Xstringext.String.has_substr s "not enough MMIO resources" -> Error Not_enough_mmio_resources
+		| e ->
+			debug "%s: set sriov error with exception %s" dev (Printexc.to_string e);
+			Error (Unknown "unable to enable via sysfs")
+
+	let get_driver dev =
```
robhoes: What is this for? We don't want any exceptions to be raised here, because the top-level IDL function returns a `result` type.

lib/network_utils.ml
```diff
@@ -1163,3 +1200,196 @@ module Ethtool = struct
...
+		match (Sysfs.get_driver_name dev) with
+		| Some driver -> driver
+		| None -> raise (Sys_error ("cannot get driver name for " ^ dev))
+
+	let write_conf_file driver content=
```
robhoes: We should have a `Modprobe` module with this and some of the functions above.

lib/network_utils.ml
```diff
@@ -1163,3 +1200,196 @@ module Ethtool = struct
...
+	a indicates the probe configuration already has the max_vfs options, meaning the device doesn't support sysfs and will be configed by modprobe
+	b indicates some changes shall be made on the coniguration to enable sriov to max_vfs, so we shall rebuild the initrd.
+	and c is the configurations after these changes
+	*)
+	let parse_modprobe_conf driver max_vfs =
+		let file_path = Printf.sprintf "/etc/modprobe.d/%s.conf" driver in
+		let option = gen_options_for_maxvfs driver max_vfs in
+		parse_modprobe_conf_internal file_path driver option
+
+	let get_sriov_info lspci_output =
```
robhoes: I think it is better to get this information (max/num_vfs) from sysfs, which is more efficient (quick file reads).

lib/network_utils.ml
```diff
@@ -1163,3 +1200,196 @@ module Ethtool = struct
...
+		if Re.execp (Re_perl.compile_pat "Capabilities:.+SR-IOV") lspci_output then
+		begin
+			let group = Re.exec_opt (Re_perl.compile_pat "Total VFs: (\d+), Number of VFs: (\d+)") lspci_output in
+			match group with
+			| None -> {max_vfs=0; num_vfs=0}
```
robhoes: Using a record for returning to simple integers is a bit overkill. A tuple is fine. However, in this case I had expected two functions: `Sysfs.sriov_numvfs` and `Sysfs.sriov_totalvfs`.

lib/network_utils.ml
```diff
@@ -1163,3 +1200,196 @@ module Ethtool = struct
...
+		call_script lspci ["-vv"; "-s"; (Sysfs.get_pcibuspath dev)]
+
+	let get_capabilities dev = 
+		try
+			let info = get_lspci_output dev |> get_sriov_info in
+			if info.max_vfs = 0 then [] else ["sriov"]
+		with _ -> []
+
+	let set_sriov_via_sysfs dev num_vfs =
+		let interface = Printf.sprintf "/sys/class/net/%s/device/sriov_numvfs" dev in
```
robhoes: Oh, so here we are using the sysfs node.

robhoes: This really deserves a `set_sriov_numvfs` function in the `Sysfs` module.

lib/network_utils.ml
```diff
@@ -1163,3 +1200,196 @@ module Ethtool = struct
...
+				let has_probe_conf, need_rebuild_intrd, conf = parse_modprobe_conf driver 0 in
+				match has_probe_conf,need_rebuild_intrd with
+				| false, false -> set_sriov_via_sysfs dev 0; ()
+				| true, true ->
+					write_conf_file driver conf;
+					rebuild_initrd ()
+				| _ -> ()
+			with e -> debug "%s: best effort disable sriov with exn: %s" dev (Printexc.to_string e);()
+		in
+		best_effort_disable(); Ok Disable_successful
```
robhoes: So the only possible return value of `disable` is `Ok Disable_successful`. That's a bit odd, isn't it?

lib/network_utils.ml
robhoes: Or, it may be easier to get the information from sysfs. For example, to check whether a NIC supports SR-IOV, we can look at the `sriov_totalvfs` node instead of the capabilities from lspci:

>What:		/sys/bus/pci/devices/.../sriov_totalvfs
Date:		November 2012
Contact:	Donald Dutile <ddutile@redhat.com>
Description:
		This file appears when a physical PCIe device supports SR-IOV.
		Userspace applications can read this file to determine the
		maximum number of Virtual Functions (VFs) a PCIe physical
		function (PF) can support. Typically, this is the value reported
		in the PF's SR-IOV extended capability structure's TotalVFs
		element.  Drivers have the ability at probe time to reduce the
		value read from this file via the pci_sriov_set_totalvfs()
		function.

xiewei20082008: I remember that I didn't use sriov_totalvfs file is because it doesn't always show the correct totalvfs number. The standard should be sriov_totalvfs = sriov_maxvfs + 1 (totalvfs = pf number + vf number), but for igbxe devices, the value in file sriov_totalvfs is 63 which is exactly the same with max_vfs. 

I used lspci's output is because its output is always correct.

xiewei20082008: Moreover, I didn't find ocaml-pci able to get the capabilities fields of PCI.

I will change to use sysfs to get sriov-totalvfs, the disadvantage of which is we will only enable 62 VFs on igbxe devices (maximum VFs is actually 63).

robhoes: Are you saying that lspci has a special case hardcoded for igbxe? Because, as far as I understand it, lspci/libpci by default also uses sysfs to get this kind of information.

For the capability checking, however, it doesn't matter whether it gets the number right. If the file exists, then the SR-IOV is present, if not, then it isn't.

xiewei20082008:
Since a lot of changes have been made, I am closing the PR and opening a new One.

PR #125, opened 2018-01-17
CP-26333 Add PIF.pci field for network sriov feature

krizex:
Implement interface to get pci bus path


robhoes:
The code is good, but the commit message doesn't make a lot of sense in this context: it should just describe the change that is made in this patch.

krizex:
Updated commit message.

PR #126, opened 2018-01-17
Mainly I changed to use Rresult for all error handle.

xiewei20082008:
Since a lot of changed have been made, I raised a new PR and closed the old one. Mainly I changed to use Rresult for all error handle.

I suppose the current sriov_error defined in IDL is too complex.
Currently:

```
type sriov_error =
	| Device_not_found
	| Bus_out_of_range
	| Not_enough_mmio_resources
	| Unknown of string
```

But actually XAPI will *not* match different tags with different behaviors. So I changed it to just

```
type sriov_error =
	| Msg of string
```

## Changed list:

moved functions related to ip link to module ip.
Refined function `parent_device_of_vf`

- directly use 'devices.(0)'
- use result type instead of exception

Refined function `device_index_of_vf`

- use result type instead of exception

Add new Dracut module

Get max/num vfs from sysfs with two functions:

- Sysfs.sriov_numvfs
- Sysfs.sriov_totalvfs

Add set_sriov_numvfs into Sysfs module.

Remove get_driver function

Add Modprobe module

Disable SRIOV will return a Rresult type

Remove `open Xcp_pci`

## Unchanged list:

I tried to remove Sriov module in Network_utils, but failed as

- The functions, `parse_modprobe_conf` and `gen_options_for_maxvfs` can only be used for enabling SRIOV - they are not general Modprobe functions, so it looks not reasonable to move the two functions into Module Modprobe
- `enable_internal` depends heavily on the error I defined in `network_utils.ml`, so I am not sure if it is appropriate to move the whole function into `network_server.ml`

I haven't yet split the commit into small commits. Can I split it after the code are ready to merge.

---
This change is [Reviewable](https://reviewable.io/reviews/xapi-project/xcp-networkd/126)

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -46,6 +46,9 @@ let brctl = ref "/sbin/brctl"
 let modprobe = "/sbin/modprobe"
 let ethtool = ref "/sbin/ethtool"
 let bonding_dir = "/proc/net/bonding/"
+let uname = ref "/usr/bin/uname"
+let dracut = ref "/sbin/dracut"
+let dracut_timeout = ref 120.0
 let fcoedriver = ref "/opt/xensource/libexec/fcoe_driver"
 let inject_igmp_query_script = ref "/usr/libexec/xenopsd/igmp_query_injector.py"
 let mac_table_size = ref 10000
@@ -1244,4 +1247,25 @@ module Ethtool = struct
 	let set_offload name options =
 		if options <> [] then
 			ignore (call ~log:true ("-K" :: name :: (List.concat (List.map (fun (k, v) -> [k; v]) options))))
+end
+
+module Dracut = struct
+	let call ?(log=false) args =
+		call_script ~timeout:(Some !dracut_timeout) ~log_successful_output:log !dracut args
+
+	let rebuild_initrd () =
+		try
+			info "Building initrd...";
+			let img_name = call_script !uname ["-r"] |> String.trim in
+			call ["-f"; Printf.sprintf "/boot/initrd-%s.img" img_name; img_name];
+			Result.Ok ()
+		with _ -> Result.Error (Fail_to_rebuild_initrd, "Error occurs in building initrd")
+end
+
+module Modprobe = struct
+	let write_conf_file driver content=
+		try
+			Unixext.write_string_to_file (Printf.sprintf "/etc/modprobe.d/%s.conf" driver) (String.concat "\n" content);
+			Result.Ok ()
+		with _ -> Result.Error (Fail_to_write_modprobe_cfg, "Failed to write modprobe configuration file for: " ^ driver)
 end
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -481,6 +481,22 @@ info "Found at [ %s ]" (String.concat ", " (List.map string_of_int indices));
 	let destroy_vlan name =
 		if List.mem name (Sysfs.list ()) then
 			ignore (call ~log:true ["link"; "delete"; name])
+
+	let set_vf_mac dev index mac =
+		try
+			Result.Ok (link_set dev ["vf"; string_of_int index; "mac"; mac])
+		with _ -> Result.Error (Fail_to_set_vf_mac, "Failed to set vf mac for: " ^ dev)
+
+	let set_vf_vlan dev index vlan =
+		try
+			Result.Ok (link_set dev ["vf"; string_of_int index; "vlan"; string_of_int vlan])
+		with _ -> Result.Error (Fail_to_set_vf_vlan, "Failed to set vf vlan for: " ^ dev)
+
+	(* We know some NICs do not support config VF Rate, so will explicitly tell XAPI this error*)
+	let set_vf_rate dev index rate =
+		try
+			Result.Ok (link_set dev ["vf"; string_of_int index; "mac"; string_of_int rate])
+		with _ -> Result.Error (Fail_to_set_vf_rate, "Failed to set vf rate for: " ^ dev)
 end
 
 module Linux_bonding = struct
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -13,11 +13,26 @@
  *)
 
 open Xapi_stdext_pervasives
-
+open Xapi_stdext_unix
+open Xapi_stdext_std
 open Network_interface
 
 module D = Debug.Make(struct let name = "network_utils" end)
 open D
+
+type util_error =
+| Bus_out_of_range
+| Not_enough_mmio_resources
+| Fail_to_set_vf_rate
+| Fail_to_set_vf_vlan
+| Fail_to_set_vf_mac
+| Parent_device_of_vf_not_found
+| Vf_index_not_found
+| Fail_to_rebuild_initrd
+| Fail_to_write_modprobe_cfg
+| Fail_to_get_driver_name
+| No_sriov_capability
+| Other
 
 let iproute2 = "/sbin/ip"
 let resolv_conf = "/etc/resolv.conf"
@@ -164,6 +179,11 @@ module Sysfs = struct
 			debug "%s: could not read netdev's driver name" dev;
 			None
 
+	let get_driver_name_err dev =
+		match get_driver_name dev with
+		| Some a -> Result.Ok a
+		| None -> Result.Error (Fail_to_get_driver_name, "Failed to get driver name for: "^ dev)
+
 	(** Returns the features bitmap for the driver for [dev].
 	 *  The features bitmap is a set of NETIF_F_ flags supported by its driver. *)
 	let get_features dev =
@@ -208,6 +228,68 @@ module Sysfs = struct
 		|> (fun p -> try read_one_line p |> duplex_of_string with _ -> Duplex_unknown)
 		in (speed, duplex)
 
+	let get_dev_nums_with_same_driver driver = 
+		try
+			Sys.readdir ("/sys/bus/pci/drivers/" ^ driver)
+			|> Array.to_list
+			|> List.filter (Re.execp (Re_perl.compile_pat "\d+:\d+:\d+\.\d+"))
+			|> List.length
+		with _ -> 0
+
+	let parent_device_of_vf pcibuspath =
+		try
+			let pf_net_path = Printf.sprintf "/sys/bus/pci/devices/%s/physfn/net" pcibuspath in
+			let devices = Sys.readdir pf_net_path in
+			Result.Ok devices.(0)
+		with _ -> Result.Error (Parent_device_of_vf_not_found, "Can not get parent device for " ^ pcibuspath)
+
+	let device_index_of_vf parent_device pcibuspath =
+		try
+			let re = Re_perl.compile_pat "virtfn(\d+)" in
+			let device_path = getpath parent_device "device" in
+			let group = Sys.readdir device_path
+				|> Array.to_list
+				|> List.filter (Re.execp re) (* List elements are like "virtfn1" *)
+				|> List.find (fun x -> Xstringext.String.has_substr (Unix.readlink (device_path ^ "/" ^ x)) pcibuspath )
+				|> Re.exec_opt re
+			in
+			match group with
+			| None -> Result.Error (Vf_index_not_found, "Can not get device index for " ^ pcibuspath)
+			| Some x -> Ok (int_of_string (Re.Group.get x 1))
+		with _ -> Result.Error (Vf_index_not_found, "Can not get device index for " ^ pcibuspath)
+
+	let get_sriov_numvfs dev =
+		try
+			getpath dev "device/sriov_numvfs"
+			|> read_one_line  
+			|> String.trim
+			|> int_of_string
+		with _ -> 0
+
+	let get_sriov_maxvfs dev =
+		try
+			getpath dev "device/sriov_totalvfs"
+			|> read_one_line  
+			|> String.trim
+			|> int_of_string
+			|> fun n -> n - 1 (* maxvfs is totalvfs -1, as totalvfs is PF num + VF num *)
+		with _ -> 0
+
+	let set_sriov_numvfs dev num_vfs =
+		let interface = getpath dev "device/sriov_numvfs" in
+		let oc = open_out interface in
+		try
+			write_one_line interface (string_of_int num_vfs);
+			if get_sriov_numvfs dev = num_vfs then Result.Ok  ()
+			else Result.Error (Other, "Error: set SR-IOV error on " ^ dev)
+		with
+		| Sys_error s when Xstringext.String.has_substr s "out of range of" -> 
+			Result.Error (Bus_out_of_range, "Error: bus out of range when setting SR-IOV numvfs on " ^ dev)
+		| Sys_error s when Xstringext.String.has_substr s "not enough MMIO resources" -> 
+			Result.Error (Not_enough_mmio_resources, "Error: not enough mmio resources when setting SR-IOV numvfs on " ^ dev)
+		| e ->
+			let msg = Printf.sprintf "Error: set SR-IOV numvfs error with exception %s on %s" (Printexc.to_string e) dev in
+			Result.Error (Other, msg)
 end
 
 module Ip = struct
--- a/networkd/network_server.ml
+++ b/networkd/network_server.ml
@@ -14,6 +14,9 @@
 
 open Network_utils
 open Network_interface
+open Xapi_stdext_std
+open Xapi_stdext_unix
+open Xapi_stdext_monadic
 
 module D = Debug.Make(struct let name = "network_server" end)
 open D
@@ -123,6 +126,125 @@ let need_enic_workaround () =
 module Sriov = struct
 	let get_capabilities dev = 
 		if Sysfs.get_sriov_maxvfs dev = 0 then [] else ["sriov"]
+
+	(* To enable SR-IOV via modprobe configuration, we add a line like `options igb max_vfs=7,7,7`		
+	into the configuration. This function is meant to generate the options like `7,7,7`. the `7` have to 		
+	be repeated as many times as the number of devices with the same driver.		
+	*)
+	let gen_options_for_maxvfs driver max_vfs =
+		match Sysfs.get_dev_nums_with_same_driver driver with
+		| num when num > 0 -> Result.Ok (
+			Array.make num (string_of_int max_vfs)
+			|> Array.to_list
+			|> String.concat ",")
+		| _ -> Result.Error (Other, "Fail to generate options for maxvfs for " ^ driver)
+
+	(* For given driver like igb, we parse each line of igb.conf which is the modprobe
+	configuration for igb. We keep the same the lines that do not have SR-IOV configurations and 
+	change lines that need to be changed with patterns like `options igb max_vfs=4`
+	*)
+	let parse_modprobe_conf_internal file_path driver option =
+		let has_probe_conf = ref false in
+		let need_rebuild_initrd = ref false in
+		let parse_single_line s = 
+			let parse_driver_options s = 
+				match Xstringext.String.split ~limit:2 '=' s with
+				(* has SR-IOV configuration but the max_vfs is exactly what we want to set, so no changes and return s *)
+				| [k; v] when k = "max_vfs" && v = option ->  has_probe_conf := true; s
+				(* has SR-IOV configuration and we need change it to expected option *)
+				| [k; v] when k = "max_vfs"  -> 
+					has_probe_conf := true;
+					need_rebuild_initrd := true;
+					debug "change SR-IOV options from [%s=%s] to [%s=%s]" k v k option;
+					Printf.sprintf "max_vfs=%s" option
+				(* we do not care the lines without SR-IOV configurations *)
+				| _ -> s
+			in
+			let trimed_s = String.trim s in
+			if Re.execp (Re_perl.compile_pat ("options[ \t]+" ^ driver)) trimed_s then
+				let driver_options = Re.split (Re_perl.compile_pat "[ \t]+") trimed_s in
+				List.map parse_driver_options driver_options
+				|> String.concat " "
+			else
+				trimed_s
+		in
+		let lines = try Unixext.read_lines file_path with _ -> [] in
+		let new_conf = List.map parse_single_line lines in
+		!has_probe_conf, !need_rebuild_initrd, new_conf
+
+	(*
+	returns ( a * b * c) where
+	a indicates the probe configuration already has the max_vfs options, meaning the device doesn't support sysfs and will be configed by modprobe
+	b indicates some changes shall be made on the coniguration to enable SR-IOV to max_vfs, so we shall rebuild the initrd.
+	and c is the configurations after these changes
+	*)
+	let parse_modprobe_conf driver max_vfs =
+		try
+			let open Rresult.R.Infix in
+			let file_path = Printf.sprintf "/etc/modprobe.d/%s.conf" driver in
+			gen_options_for_maxvfs driver max_vfs >>= fun options ->
+			Result.Ok (parse_modprobe_conf_internal file_path driver options)
+		with _ -> Result.Error (Other, "Failed to parse modprobe conf for SR-IOV configuration for " ^ driver)
+
+	let enable_sriov_via_modprobe driver maxvfs has_probe_conf need_rebuild_initrd conf = 
+		let open Rresult.R.Infix in
+		match has_probe_conf, need_rebuild_initrd with
+		| true, true ->
+			Modprobe.write_conf_file driver conf >>= fun () ->
+			Dracut.rebuild_initrd ()
+		| false, false -> 
+			gen_options_for_maxvfs driver maxvfs >>= fun options ->
+			let new_option_line = Printf.sprintf "options %s max_vfs=%s" driver options in
+			Modprobe.write_conf_file driver (conf @ [new_option_line]) >>= fun () ->
+			Dracut.rebuild_initrd ()
+		| true, false -> Result.Ok () (* already have modprobe configuration and no need to change *)
+		| false, true -> Result.Error (Other, "enabling SR-IOV via modprobe never comes here for: " ^ driver)
+
+	let enable_internal dev =
+		let open Rresult.R.Infix in
+		let numvfs = Sysfs.get_sriov_numvfs dev
+		and maxvfs = Sysfs.get_sriov_maxvfs dev in
+		Sysfs.get_driver_name_err dev >>= fun driver ->
+		parse_modprobe_conf driver maxvfs >>= fun (has_probe_conf, need_rebuild_initrd, conf) ->
+		if maxvfs = 0 then Result.Error (No_sriov_capability, (Printf.sprintf "%s: do not have SR-IOV capabilities" dev)) else Ok () >>= fun () ->
+		(* We cannot first call sysfs method unconditionally for the case where the `SR-IOV hardware status` is ON and 
+		we are about to enable SR-IOV, which is the else case. It is because sysfs method to enable SR-IOV on a `SR-IOV 
+		already enabled device` will always be succuessful even if the device doesn't support sysfs. A simple example is as follows:
+
+		- a device that doesn't support sysfs
+		- enable the device via modprobe -> Hardware Status OFF
+		- reboot -> Status ON
+		- disable via modprobe before reboot -> Status ON
+		- enable via sysfs -> will return Sysfs_succesfully -> Here bug arise.*)
+		if numvfs = 0 then begin
+			debug "enable SR-IOV on a device: %s that is disabled" dev;
+			match Sysfs.set_sriov_numvfs dev maxvfs with
+			| Result.Ok _ -> Ok Sysfs_successful
+			| Result.Error (Bus_out_of_range, msg) as e ->
+				debug "%s" msg; e
+			| Result.Error (Not_enough_mmio_resources, msg) as e ->
+				debug "%s" msg; e
+			| Result.Error (_, msg) ->
+				debug "%s does not support sysfs interfaces for reason %s, trying modprobe" dev msg;
+				enable_sriov_via_modprobe driver maxvfs has_probe_conf need_rebuild_initrd conf >>= fun () ->
+				Ok Modprobe_successful_requires_reboot
+		end
+		else begin
+			debug "enable SR-IOV on a device: %s that has been already enabled" dev;
+			match has_probe_conf with
+			| false -> Ok Sysfs_successful
+			| true -> 
+				enable_sriov_via_modprobe driver maxvfs has_probe_conf need_rebuild_initrd conf >>= fun () ->
+				Ok Modprobe_successful
+		end
+
+	let enable _ dbg ~name =
+		Debug.with_thread_associated dbg (fun () ->
+			debug "Enable network SR-IOV by name: %s" name;
+			match enable_internal name with
+			| Ok t -> (Ok t:enable_result)
+			| Result.Error (_, msg) -> Error msg
+		) ()
 end
 
 module Interface = struct
--- a/networkd/network_server.ml
+++ b/networkd/network_server.ml
@@ -245,6 +245,28 @@ module Sriov = struct
 			| Ok t -> (Ok t:enable_result)
 			| Result.Error (_, msg) -> Error msg
 		) ()
+
+	let disable_internal dev =
+		let open Rresult.R.Infix in
+		Sysfs.get_driver_name_err dev >>= 
+		fun driver ->
+		parse_modprobe_conf driver 0 >>= 
+		fun (has_probe_conf, need_rebuild_intrd, conf) ->
+		match has_probe_conf,need_rebuild_intrd with
+		| false, false ->
+			Sysfs.set_sriov_numvfs dev 0
+		| true, true ->
+			Modprobe.write_conf_file driver conf >>= fun () ->
+			Dracut.rebuild_initrd ()
+		| _ -> Ok ()
+
+	let disable _ dbg ~name =
+		Debug.with_thread_associated dbg (fun () ->	
+			debug "Disable network SR-IOV by name: %s" name;
+			match disable_internal name with
+			| Ok () -> (Ok:disable_result)
+			| Result.Error (_, msg) -> Error msg
+		) ()
 end
 
 module Interface = struct
--- a/networkd/network_server.ml
+++ b/networkd/network_server.ml
@@ -267,6 +267,32 @@ module Sriov = struct
 			| Ok () -> (Ok:disable_result)
 			| Result.Error (_, msg) -> Error msg
 		) ()
+
+	let make_vf_conf_internal pcibuspath mac vlan rate =
+		let exe_except_none f = function
+			| None -> Result.Ok ()
+			| Some a -> f a
+		in
+		let open Rresult.R.Infix in
+		Sysfs.parent_device_of_vf pcibuspath >>= fun dev ->
+		Sysfs.device_index_of_vf dev pcibuspath >>= fun index ->
+		exe_except_none (Ip.set_vf_mac dev index) mac >>= fun () ->
+		exe_except_none (Ip.set_vf_vlan dev index) vlan >>= fun () ->
+		exe_except_none (Ip.set_vf_rate dev index) rate
+
+	let make_vf_config _ dbg ~pci_address ~(vf_info : Sriov.sriov_pci_t)=
+		Debug.with_thread_associated dbg (fun () ->	
+			let vlan = Opt.map Int64.to_int vf_info.vlan
+			and rate = Opt.map Int64.to_int vf_info.rate
+			and pcibuspath = Xcp_pci.string_of_address pci_address in
+			debug "Config VF with pci address: %s" pcibuspath;
+			match make_vf_conf_internal pcibuspath vf_info.mac vlan rate with
+			| Result.Ok () -> (Ok:config_result)
+			| Result.Error (Fail_to_set_vf_rate, msg) -> 
+				debug "%s" msg;
+				Error Config_vf_rate_not_supported
+			| Result.Error (_, msg) -> debug "%s" msg; Error (Unknown msg)
+		) ()
 end
 
 module Interface = struct
--- a/networkd/network_server.ml
+++ b/networkd/network_server.ml
@@ -120,6 +120,11 @@ let need_enic_workaround () =
 		| Some vs -> (is_older_version vs !enic_workaround_until_version ())
 		| None -> false )
 
+module Sriov = struct
+	let get_capabilities dev = 
+		if Sysfs.get_sriov_maxvfs dev = 0 then [] else ["sriov"]
+end
+
 module Interface = struct
 	let get_config name =
 		get_config !config.interface_config default_interface name
@@ -374,7 +379,7 @@ module Interface = struct
 
 	let get_capabilities _ dbg ~name =
 		Debug.with_thread_associated dbg (fun () ->
-			Fcoe.get_capabilities name
+			Fcoe.get_capabilities name @ Sriov.get_capabilities name
 		) ()
 
 	let is_connected _ dbg ~name =
@@ -999,28 +1004,6 @@ module Bridge = struct
 					) ports
 				)
 			) config
-		) ()
-end
-
-module Sriov = struct
-	open Xcp_pci
-
-	let enable _ dbg ~name =
-		Debug.with_thread_associated dbg (fun () ->
-			debug "Enable NET-SRIOV by name: %s" name;
-			Ok Modprobe_successful
-		) ()
-
-	let disable _ dbg ~name =
-		Debug.with_thread_associated dbg (fun () ->	
-			debug "Disable NET-SRIOV by name: %s" name;
-			Ok Modprobe_successful
-		) ()
-
-	let make_vf_config _ dbg ~pci_address ~vf_info =
-		Debug.with_thread_associated dbg (fun () ->	
-			let pcibuspath = string_of_address pci_address in
-			debug "Config VF with pci address: %s" pcibuspath;
 		) ()
 end
 
--- a/lib/jbuild
+++ b/lib/jbuild
@@ -8,6 +8,7 @@
               rpclib
               systemd
               threads
+              re.perl
               xapi-stdext-unix
               xcp-inventory
               xcp.network))
--- a/networkd/networkd.ml
+++ b/networkd/networkd.ml
@@ -61,6 +61,9 @@ let options = [
 	"igmp-query-maxresp-time", Arg.Set_string Network_utils.igmp_query_maxresp_time, (fun () -> !Network_utils.igmp_query_maxresp_time), "Maximum Response Time in IGMP Query message to send";
 	"enable-ipv6-mcast-snooping", Arg.Bool (fun x -> Network_utils.enable_ipv6_mcast_snooping := x), (fun () -> string_of_bool !Network_utils.enable_ipv6_mcast_snooping), "IPv6 multicast snooping toggle";
 	"mcast-snooping-disable-flood-unregistered", Arg.Bool (fun x -> Network_utils.mcast_snooping_disable_flood_unregistered := x), (fun () -> string_of_bool !Network_utils.mcast_snooping_disable_flood_unregistered), "Set OVS bridge configuration mcast-snooping-disable-flood-unregistered as 'true' or 'false'";
+	"uname_cmd_path", Arg.Set_string Network_utils.uname, (fun () -> !Network_utils.uname), "Path to the Unix command uname";
+	"dracut_cmd_path", Arg.Set_string Network_utils.dracut, (fun () -> !Network_utils.dracut), "Path to the Unix command dracut";
+	"dracut_timeout", Arg.Set_float Network_utils.dracut_timeout, (fun () -> string_of_float !Network_utils.dracut_timeout), "Default value for the dracut command timeout";
 ]
 
 let start server =


networkd/network_server.ml
```diff
@@ -999,24 +1001,43 @@ module Bridge = struct
...
+		exe_except_none (Ip.set_vf_rate dev index) rate >>= fun () ->
+		Ok ()
+
 	let make_vf_config _ dbg ~pci_address ~vf_info =
 		Debug.with_thread_associated dbg (fun () ->	
-			let pcibuspath = string_of_address pci_address in
+			let pcibuspath = Xcp_pci.string_of_address pci_address in
 			debug "Config VF with pci address: %s" pcibuspath;
+			make_vf_conf_internal pcibuspath vf_info;
+			()
```
robhoes: `make_vf_conf_internal` returns a `result`, but this function throws the return value away and aways returns `()`. I would expect something like this:
```
match make_vf_conf_internal pcibuspath vf_info with
| Ok () -> ()
| Error e -> (* raise some exception *)
```
Or even change the return type of `make_vf_conf` in the the IDL to be a `result` type, and possibly translate the result types in a match as above.

lib/network_utils.ml
```diff
@@ -208,6 +230,72 @@ module Sysfs = struct
 		|> (fun p -> try read_one_line p |> duplex_of_string with _ -> Duplex_unknown)
 		in (speed, duplex)
 
+	let get_dev_nums_with_same_driver driver = 
+		try
+			Sys.readdir ("/sys/bus/pci/drivers/" ^ driver)
+			|> Array.to_list
+			|> List.filter (Re.execp (Re_perl.compile_pat "\d+:\d+:\d+\.\d+"))
```
mseri: Can't we use `Scanf` instead of introducing a new dependency? I remember that some `Re` module had issues with thread safety

networkd/network_server.ml
xiewei20082008: Ok. I will change the return type of make_vf_conf in the IDL to be a result to keep the consistency with the enable/disable.

xiewei20082008:
*[lib/network_utils.ml, line 237 at r1](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3BfnhsEY6iGY96dm42-r1-237:-L3CneFMAo42K3Yw4jUg:b-inph8c) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/da1f0ea67d0aa0d0fef138fdd9c094cf6bb23501/lib/network_utils.ml#L237)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

Can't we use `Scanf` instead of introducing a new dependency? I remember that some `Re` module had issues with thread safety. The same applies below.
</blockquote></details>

- I ever tried to use Scanf, but that makes the code not as simple as regex - for Scanf a try-catch expressions have to be involved to check if a pattern match successfully.
- I asked this question also in Ring3 channel, looks like it is the Re_str in std lib not thread-safe, but not the Re module. Furthermore, our current toolstack code have a lot of Re_str from the Re module already in use.
- In line 1330, anyway Re.split will be used to split the string.

As above, Re module dependency is not newly introduced to Toolstack, it is just perl style regex introduced, so I hope I can keep use Re_perl in the code.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

lib/network_utils.ml
```diff
@@ -208,6 +230,72 @@ module Sysfs = struct
...
+			|> int_of_string
+		with _ -> 0
+
+	let get_sriov_maxvfs dev =
+		try
+			getpath dev "device/sriov_totalvfs"
+			|> read_one_line  
+			|> String.trim
+			|> int_of_string
+			|> (-) 1 (* maxvfs is totalvfs -1, as totalvfs is PF num + VF num *)
```
lindig: This computes `1 - sriov_totalvfs`. You probably want `sriov_totalvfs - 1`. Use `|> fun n -> n - 1`

xiewei20082008:
*[networkd/network_server.ml, line 1040 at r1](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3BfnhsEY6iGY96dm43-r1-1040:-L3SdvewrlrBm82tZ64C:b-896fix) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/da1f0ea67d0aa0d0fef138fdd9c094cf6bb23501/networkd/network_server.ml#L1040)):*
<details><summary><i>Previously, xiewei20082008 (Wei Xie) wrote…</i></summary><blockquote>

Ok. I will change the return type of make_vf_conf in the IDL to be a result to keep the consistency with the enable/disable.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

xiewei20082008:
*[lib/network_utils.ml, line 277 at r1](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3BfnhsEY6iGY96dm42-r1-277:-L3Se-8VjCRwOmB6yhna:b-896fix) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/da1f0ea67d0aa0d0fef138fdd9c094cf6bb23501/lib/network_utils.ml#L277)):*
<details><summary><i>Previously, lindig (Christian Lindig) wrote…</i></summary><blockquote>

This computes `1 - sriov_totalvfs`. You probably want `sriov_totalvfs - 1`. Use `|> fun n -> n - 1`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

mseri:
*[lib/network_utils.ml, line 19 at r2](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3ShTkjzyNBpaOz3ajS:-L3ShTkjzyNBpaOz3ajT:bvsv48c) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/f5d7fe7450dffc52bc65c25801df72f93d0598b1/lib/network_utils.ml#L19)):*
> ```OCaml
> open Xapi_stdext_std
> open Network_interface
> open Rresult.R.Infix
> ```

Here as well, let's open `Rresult.R.Infix` inline where we use it instead of the toplevel

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

xiewei20082008:
*[networkd/network_server.ml, line 17 at r1](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3Sg2DeSxtZtFblXnmc:-L3W9tIhWByiT-7ixXRn:b-896fix) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/da1f0ea67d0aa0d0fef138fdd9c094cf6bb23501/networkd/network_server.ml#L17)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

I think it's better to open this locally where actually used and not in the toplevel.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

xiewei20082008:
*[lib/network_utils.ml, line 19 at r2](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3ShTkjzyNBpaOz3ajS:-L3WAA1hXdcy-WOEAF80:b-896fix) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/f5d7fe7450dffc52bc65c25801df72f93d0598b1/lib/network_utils.ml#L19)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

Here as well, let's open `Rresult.R.Infix` inline where we use it instead of the toplevel
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

robhoes:
Review status: 0 of 3 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[lib/network_utils.ml, line 1408 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3YID1YngHnLfhlx9Cv:-L3YID1YngHnLfhlx9Cw:bvnscwq) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1408)):*
> ```OCaml
> 		| _ -> Ok ()
> 
> 	let make_vf_conf_internal pcibuspath mac vlan rate =
> ```

I think that this function should just be inline in `make_vf_conf` in network_server.ml. I don't really like the `*_internal` functions here. This network_utils module is meant to be a collections of simple functions around Linux/system tools that are sequenced from network_server. This function already uses  only calls to other modules/functions of network_util, so it can easily be moved.

---

*[lib/network_utils.ml, line 1419 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3YJMdEw5YMR8L35kSg:-L3YJMdEw5YMR8L35kSh:b-zc0bd6) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1419)):*
> ```OCaml
> 		exe_except_none (Ip.set_vf_vlan dev index) vlan >>= fun () ->
> 		exe_except_none (Ip.set_vf_rate dev index) rate >>= fun () ->
> 		Result.Ok ()
> ```

You don't really need this last line, as the one above will already return `Ok` if all is well. I do like the way this looks now with the use of the error monad!

---

*[networkd/network_server.ml, line 378 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3Xf2sF6fvKh_l2fmpB:-L3Xf2sF6fvKh_l2fmpC:b-9d42gc) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/networkd/network_server.ml#L378)):*
> ```OCaml
> 	let get_capabilities _ dbg ~name =
> 		Debug.with_thread_associated dbg (fun () ->
> 			Fcoe.get_capabilities name @ Network_utils.Sriov.get_capabilities name
> ```

You don't actually need `Network_utils.` because this _is_ opened on top (below as well).

---

*[networkd/network_server.ml, line 1024 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3XfeFTi6i12rHabsKO:-L3XfeFUQKoiPxqrRoQl:bqmz4es) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/networkd/network_server.ml#L1024)):*
> ```OCaml
> 		) ()
> 
> 	let make_vf_config _ dbg ~pci_address ~(vf_info : Sriov.sriov_pci_t)=
> ```

I don't think that this type annotation add much value?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

robhoes:
Review status: 0 of 3 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[lib/network_utils.ml, line 41 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3YYg8kejEB5Mmf9IrN:-L3YYg8kejEB5Mmf9IrO:benusqi) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L41)):*
> ```OCaml
> let ethtool = ref "/sbin/ethtool"
> let bonding_dir = "/proc/net/bonding/"
> let uname = "/usr/bin/uname"
> ```

Please make all of these `ref`s so that they can be set from a config file. Code needs to be added to networkd.ml to wire this up.

---

*[lib/network_utils.ml, line 1348 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3Y_3qPuavvbwm3Br50:-L3Y_3qPuavvbwm3Br51:b2igbng) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1348)):*
> ```OCaml
> 		with _ -> Result.Error (Other, "Failed to parse modprobe conf for SRIOV configuration for " ^ driver)
> 
> 	let get_capabilities dev = 
> ```

See note on `make_vf_conf_internal` below regarding the split between network_server and network_utils. I'd move this into network_server.

---

*[lib/network_utils.ml, line 1351 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3Y_Q4HLr9cXhJiOLFR:-L3Y_Q4I8IVEXJpz_kxI:bh7mt0m) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1351)):*
> ```OCaml
> 		if Sysfs.get_sriov_maxvfs dev = 0 then [] else ["sriov"]
> 
> 	let enable_internal dev =
> ```

Again, see note on `make_vf_conf_internal` below regarding the split between network_server and network_utils. I'd move this into network_server.

---

*[lib/network_utils.ml, line 1356 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3YaJN6PgF6baVgL9cg:-L3YaJN6PgF6baVgL9ch:btychu4) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1356)):*
> ```OCaml
> 		and maxvfs = Sysfs.get_sriov_maxvfs dev in
> 		Sysfs.get_driver_name_err dev >>= 
> 		fun driver ->
> ```

We'd normally put this on the previous line.

---

*[lib/network_utils.ml, line 1359 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3YdEaDzytjwyyRqqvT:-L3YdEaDzytjwyyRqqvU:b-7f6ct8) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1359)):*
> ```OCaml
> 		parse_modprobe_conf driver maxvfs >>=
> 		fun (has_probe_conf, need_rebuild_initrd, conf) ->
> 		let enable_sriov_via_modprobe ()=
> ```

I think that this whole function is quite long and the control flow can be made clearer, starting by taking this function outside of the main one (replace the unit arg by the ones it actually uses).

It should become very easy to see that this function tries to enable by sysfs and uses the modprobe method as fallback.

---

*[lib/network_utils.ml, line 1371 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3Yhg_-An6C_Y1lyrhp:-L3Yhg_-An6C_Y1lyrhq:bcpyimd) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1371)):*
> ```OCaml
> 			| _ -> Ok ()
> 		in
> 		if maxvfs = 0 then Result.Error (Other, (Printf.sprintf "%s: do not have sriov capabilities" dev))
> ```

This looks like a good candidate for an explicit error code.
Also, you could consider flattening the if-statement (not sure if it is clearer):
```
if maxvfs = 0 then Error xxx else Ok () >>= fun () ->
```

---

*[lib/network_utils.ml, line 1372 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3ZwDbbQGcc1tOEJT24:-L3ZwDbbQGcc1tOEJT25:bncsmz0) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1372)):*
> ```OCaml
> 		in
> 		if maxvfs = 0 then Result.Error (Other, (Printf.sprintf "%s: do not have sriov capabilities" dev))
> 		else if numvfs = 0 then begin
> ```

This if-statement is probably relevant only for the modprobe case. I assume that we can call `Sysfs.set_sriov_numvfs` even if `numvfs <> 0`, without any bad side-effects (i.e. it is idempotent)? Then I think that it is clearer to push this down into the modprobe case below, and first to the sysfs method unconditionally.

---

*[lib/network_utils.ml, line 1394 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3Y_SmvJeUXygqjxqEc:-L3Y_SmvJeUXygqjxqEd:bh7mt0m) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1394)):*
> ```OCaml
> 		end
> 
> 	let disable_internal dev =
> ```

Again, see note on `make_vf_conf_internal` below regarding the split between network_server and network_utils. I'd move this into network_server.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

robhoes:
Reviewed 1 of 3 files at r1, 1 of 2 files at r3.
Review status: 2 of 3 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

xiewei20082008:
*[lib/network_utils.ml, line 1348 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3Y_3qPuavvbwm3Br50:-L3_oiaApMOqRbPKFoO9:b-2rhgvv) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1348)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

See note on `make_vf_conf_internal` below regarding the split between network_server and network_utils. I'd move this into network_server.
</blockquote></details>

I saw this comment last time and I wrote the reason why I didn't move them to network_server (actually I had moved but move back), but looks like they are just shown in github not Reviewable..

The reasons I didn't move them to network_server are:

- Network_server is compiled into a executable instead of a library, so the functions in Network_server is not easy to be called by other Networkd part. (e.g., What about the case where I have another Module in Network_util want to call enabling SRIOV)
- I know Network_util is just used for wrapping Linux commands, so do you think we need another `helper.ml` to include these SRIOV functions?
- Currently the `get_capabilities` function of FCoE is also in Network_utils - look like it is not reasonable to be in the Network_utils either, so if we have a helper.ml we can also put the `get_capabilities` function of FCoE in helper.ml

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

xiewei20082008:
*[lib/network_utils.ml, line 1348 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3Y_3qPuavvbwm3Br50:-L3_v_Gzf8IQXhe-Fz8b:b-tscnui) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1348)):*
<details><summary><i>Previously, xiewei20082008 (Wei Xie) wrote…</i></summary><blockquote>

I saw this comment last time and I wrote the reason why I didn't move them to network_server (actually I had moved but move back), but looks like they are just shown in github not Reviewable..

The reasons I didn't move them to network_server are:

- Network_server is compiled into a executable instead of a library, so the functions in Network_server is not easy to be called by other Networkd part. (e.g., What about the case where I have another Module in Network_util want to call enabling SRIOV)
- I know Network_util is just used for wrapping Linux commands, so do you think we need another `helper.ml` to include these SRIOV functions?
- Currently the `get_capabilities` function of FCoE is also in Network_utils - look like it is not reasonable to be in the Network_utils either, so if we have a helper.ml we can also put the `get_capabilities` function of FCoE in helper.ml
</blockquote></details>

Sorry for letting you comment again.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

xiewei20082008:
*[lib/network_utils.ml, line 41 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3YYg8kejEB5Mmf9IrN:-L3bZ4V_jOLkftk6SXlW:b-wpo7oo) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L41)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

Please make all of these `ref`s so that they can be set from a config file. Code needs to be added to networkd.ml to wire this up.
</blockquote></details>

Do I just need to refine `uname` and `dracut`? or the whole commands above that are not related to SRIOV?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

xiewei20082008:
*[networkd/network_server.ml, line 378 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3Xf2sF6fvKh_l2fmpB:-L3baAFl4BWf1cq8a5gg:b-rgbpzf) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/networkd/network_server.ml#L378)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

You don't actually need `Network_utils.` because this _is_ opened on top (below as well).
</blockquote></details>

This is because `Network_utils.Sriov` has the same name with `Network_server.Sriov` - just to make the compiler happy.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

robhoes:
Review status: 2 of 3 files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[lib/network_utils.ml, line 41 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3YYg8kejEB5Mmf9IrN:-L3biNvkg_oBq4Ra4CDo:byesyvn) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L41)):*
<details><summary><i>Previously, xiewei20082008 (Wei Xie) wrote…</i></summary><blockquote>

Do I just need to refine `uname` and `dracut`? or the whole commands above that are not related to SRIOV?
</blockquote></details>

Just the ones that you've added.

---

*[lib/network_utils.ml, line 1348 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3Y_3qPuavvbwm3Br50:-L3bfVgyhsqUbmmluM5X:b-vxp2pr) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1348)):*
<details><summary><i>Previously, xiewei20082008 (Wei Xie) wrote…</i></summary><blockquote>

Sorry for letting you comment again.
</blockquote></details>

I wouldn't worry too much about what is currently a library and what isn't. This is easy to change if we need to.

The structure of networkd is really quite simple. There are three parts:

1. Wrappers around Linux tools: network_utils. This should be as simple as possible without much added logic.
2. Higher level functionality that use those tools: network_server. Enabling and disabling SR-IOV looks to be in this category, because it is our own logic. There shouldn't be anything in network_utils that need to call this.
3. A monitor for network stats and related: network_monitor_thread. This has access to and uses _both_ of the above.

I'd like to avoid adding another layer of helper modules to keep this simple.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

robhoes:
Reviewed 1 of 2 files at r3.
Review status: all files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

*[lib/network_utils.ml, line 495 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3brMWwSdEUE0I_ZNEA:-L3brMWwSdEUE0I_ZNEB:bo0c61d) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L495)):*
> ```OCaml
> 		try
> 			Result.Ok (link_set dev ["vf"; string_of_int index; "mac"; string_of_int rate])
> 		with _ -> Result.Error (Fail_to_set_vf_rate, "Failed to set vf rate for: " ^ dev)
> ```

This one's got an explicit error. Why not do that for the other ones as well, rather than `Other "<string>"`?

---

*[lib/network_utils.ml, line 1286 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3buL90wp727WBRulzE:-L3buL90wp727WBRulzF:b6qaawd) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1286)):*
> ```OCaml
> 
> module Sriov = struct
> 	let gen_options_for_maxvfs driver max_vfs =
> ```

A comment on top would be useful.

---

*[lib/network_utils.ml, line 1296 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3btqTHaEPXxiIFAYHY:-L3btqTHaEPXxiIFAYHZ:b92wqj8) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1296)):*
> ```OCaml
> 		match Sysfs.get_dev_nums_with_same_driver driver with
> 		| num when num > 0 -> Result.Ok (
> 			gen_list (string_of_int max_vfs) num
> ```

Perhaps this is easier:

    Array.make num (string_of_int max_vfs)
    |> Array.to_list
    |> String.concat ","

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

mseri:
*[lib/network_utils.ml, line 276 at r2](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3X_9eXuZkS124uEfUC:-L3X_9eYe9ogq8pkZs5Z:bce611c) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/f5d7fe7450dffc52bc65c25801df72f93d0598b1/lib/network_utils.ml#L276)):*
> ```OCaml
> 	let set_sriov_numvfs dev num_vfs =
> 		let interface = getpath dev "device/sriov_numvfs" in
> 		let oc = open_out interface in
> ```

Why not `Unixext.write_string_to_file` as we have done in the other places?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

mseri:
*[lib/network_utils.ml, line 291 at r2](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3XZkQLN2LiJunWR2UO:-L3XZkQLN2LiJunWR2UP:b-vrapd9) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/f5d7fe7450dffc52bc65c25801df72f93d0598b1/lib/network_utils.ml#L291)):*
> ```OCaml
> 			Result.Error (Not_enough_mmio_resources, "Error: not enough mmio resources when setting SRIOV numvfs on " ^ dev)
> 		| e ->
> 			let msg = Printf.sprintf "Error: set sriov error with exception %s on %s" (Printexc.to_string e) dev in
> ```

We always use `SRIOV` full caps, let's stick with the convention here as well

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

xiewei20082008:
Review status: all files reviewed at latest revision, 19 unresolved discussions, some commit checks failed.

---

*[lib/network_utils.ml, line 276 at r2](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3X_9eXuZkS124uEfUC:-L3fPqJWGkjRQpnYbZ1j:bnwoy1n) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/f5d7fe7450dffc52bc65c25801df72f93d0598b1/lib/network_utils.ml#L276)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

Why not `Unixext.write_string_to_file` as we have done in the other places?
</blockquote></details>

We cannot do that. Sysfs interface is a file that you can just open and write something in, but not allow to be removed or replaced, while Unixext.write_string_to_file is trying to replace the file with a new one.

---

*[lib/network_utils.ml, line 495 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3brMWwSdEUE0I_ZNEA:-L3fZLNEWrwpC0CUB3lL:b-jfsvl4) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L495)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

This one's got an explicit error. Why not do that for the other ones as well, rather than `Other "<string>"`?
</blockquote></details>

For our currently supported NICs, all those NICs support MAC configuration and VLAN configuration, but some NICs do not support Rate configuration. We pass the `Fail_to_set_vf_rate` as a explicit error to XAPI to let XAPI know that this is not a general error but an error of hardware itself.

---

*[lib/network_utils.ml, line 1372 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3ZwDbbQGcc1tOEJT24:-L3gGIEmXQlBDl7OkypN:bz3fgww) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1372)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

This if-statement is probably relevant only for the modprobe case. I assume that we can call `Sysfs.set_sriov_numvfs` even if `numvfs <> 0`, without any bad side-effects (i.e. it is idempotent)? Then I think that it is clearer to push this down into the modprobe case below, and first to the sysfs method unconditionally.
</blockquote></details>

We cannot first call sysfs method unconditionally for the case where the `SRIOV hardware status` is ON and we are about to enable SRIOV, which is the else case. It is because sysfs method to enable SRIOV on a `SRIOV already enabled device` will always be succuessful even if the device doesn't support sysfs.   A simple example is as follows:

- a device that doesn't support sysfs 
- enable the device via modprobe -> Hardware Status OFF
- reboot -> Status ON
- disable via modprobe before reboot -> Status ON
- enable via sysfs -> will return Sysfs_succesfully   -> Here bug arise.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

xiewei20082008:
*[networkd/network_server.ml, line 1024 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3XfeFTi6i12rHabsKO:-L3gaFEls2Ijp05z3tth:b-l0ovz0) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/networkd/network_server.ml#L1024)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

I don't think that this type annotation add much value?
</blockquote></details>

This is just to make compiler happy, since we have another Network_interface.bridge_config_t.vf_info with the same name.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

robhoes:
Review status: 1 of 4 files reviewed at latest revision, 18 unresolved discussions, some commit checks failed.

---

*[lib/network_utils.ml, line 276 at r2](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3X_9eXuZkS124uEfUC:-L3gdHRGu5vbTMZ5kYW-:b-6szqwo) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/f5d7fe7450dffc52bc65c25801df72f93d0598b1/lib/network_utils.ml#L276)):*
<details><summary><i>Previously, xiewei20082008 (Wei Xie) wrote…</i></summary><blockquote>

We cannot do that. Sysfs interface is a file that you can just open and write something in, but not allow to be removed or replaced, while Unixext.write_string_to_file is trying to replace the file with a new one.
</blockquote></details>

And `write_one_line` from this module?

---

*[lib/network_utils.ml, line 291 at r2](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3XZkQLN2LiJunWR2UO:-L3gdNv-rqBpksN3jUhZ:b-2vwicm) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/f5d7fe7450dffc52bc65c25801df72f93d0598b1/lib/network_utils.ml#L291)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

We always use `SRIOV` full caps, let's stick with the convention here as well
</blockquote></details>

SR-IOV actually ;)

---

*[lib/network_utils.ml, line 1372 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3ZwDbbQGcc1tOEJT24:-L3gdsefiCYsVGUDvKmg:btz6npc) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1372)):*
<details><summary><i>Previously, xiewei20082008 (Wei Xie) wrote…</i></summary><blockquote>

We cannot first call sysfs method unconditionally for the case where the `SRIOV hardware status` is ON and we are about to enable SRIOV, which is the else case. It is because sysfs method to enable SRIOV on a `SRIOV already enabled device` will always be succuessful even if the device doesn't support sysfs.   A simple example is as follows:

- a device that doesn't support sysfs 
- enable the device via modprobe -> Hardware Status OFF
- reboot -> Status ON
- disable via modprobe before reboot -> Status ON
- enable via sysfs -> will return Sysfs_succesfully   -> Here bug arise.
</blockquote></details>

OK cool, it would be good to explain this in a comment in the code, otherwise someone might decide to "optimise" the code later.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

xiewei20082008:
Review status: 1 of 4 files reviewed at latest revision, 16 unresolved discussions.

---

*[lib/network_utils.ml, line 41 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3YYg8kejEB5Mmf9IrN:-L3gbZABW950pj1nz4oE:b-896fix) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L41)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

Just the ones that you've added.
</blockquote></details>

Done.

---

*[lib/network_utils.ml, line 1286 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3buL90wp727WBRulzE:-L3gimuqv-DgT0tq7O6N:b-896fix) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1286)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

A comment on top would be useful.
</blockquote></details>

Done.

---

*[lib/network_utils.ml, line 1296 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3btqTHaEPXxiIFAYHY:-L3gir-xW7f4PZm_fGQz:brn3xbw) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1296)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

Perhaps this is easier:

    Array.make num (string_of_int max_vfs)
    |> Array.to_list
    |> String.concat ","
</blockquote></details>

Done. A magic was learnt.

---

*[lib/network_utils.ml, line 1348 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3Y_3qPuavvbwm3Br50:-L3gj3x_41cCoVu5CtCn:bgkgn4a) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1348)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

I wouldn't worry too much about what is currently a library and what isn't. This is easy to change if we need to.

The structure of networkd is really quite simple. There are three parts:

1. Wrappers around Linux tools: network_utils. This should be as simple as possible without much added logic.
2. Higher level functionality that use those tools: network_server. Enabling and disabling SR-IOV looks to be in this category, because it is our own logic. There shouldn't be anything in network_utils that need to call this.
3. A monitor for network stats and related: network_monitor_thread. This has access to and uses _both_ of the above.

I'd like to avoid adding another layer of helper modules to keep this simple.
</blockquote></details>

Done. Understood.

---

*[lib/network_utils.ml, line 1351 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3Y_Q4HLr9cXhJiOLFR:-L3gj822cfmaRcV_rbMJ:b-896fix) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1351)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

Again, see note on `make_vf_conf_internal` below regarding the split between network_server and network_utils. I'd move this into network_server.
</blockquote></details>

Done.

---

*[lib/network_utils.ml, line 1356 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3YaJN6PgF6baVgL9cg:-L3gj8tPTfPm3mDX6oVO:b-896fix) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1356)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

We'd normally put this on the previous line.
</blockquote></details>

Done.

---

*[lib/network_utils.ml, line 1359 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3YdEaDzytjwyyRqqvT:-L3gjApY6aSR2LezHalB:b5t3ok5) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1359)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

I think that this whole function is quite long and the control flow can be made clearer, starting by taking this function outside of the main one (replace the unit arg by the ones it actually uses).

It should become very easy to see that this function tries to enable by sysfs and uses the modprobe method as fallback.
</blockquote></details>

Done. I think I didn't give them a good name for `has_probe_conf, need_rebuild_initrd, conf`.

---

*[lib/network_utils.ml, line 1371 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3Yhg_-An6C_Y1lyrhp:-L3gjEDISAlkq8LnCbwF:brn3xbw) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1371)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

This looks like a good candidate for an explicit error code.
Also, you could consider flattening the if-statement (not sure if it is clearer):
```
if maxvfs = 0 then Error xxx else Ok () >>= fun () ->
```
</blockquote></details>

Done. A magic was learnt.

---

*[lib/network_utils.ml, line 1394 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3Y_SmvJeUXygqjxqEc:-L3gjNrcqAsEd_HtmgQe:b-896fix) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1394)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

Again, see note on `make_vf_conf_internal` below regarding the split between network_server and network_utils. I'd move this into network_server.
</blockquote></details>

Done.

---

*[lib/network_utils.ml, line 1408 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3YID1YngHnLfhlx9Cv:-L3gjOY2z4wMpogvNsBF:b-896fix) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1408)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

I think that this function should just be inline in `make_vf_conf` in network_server.ml. I don't really like the `*_internal` functions here. This network_utils module is meant to be a collections of simple functions around Linux/system tools that are sequenced from network_server. This function already uses  only calls to other modules/functions of network_util, so it can easily be moved.
</blockquote></details>

Done.

---

*[lib/network_utils.ml, line 1419 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3YJMdEw5YMR8L35kSg:-L3gjPUWnHivnC8Xf3wZ:brn3xbw) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1419)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

You don't really need this last line, as the one above will already return `Ok` if all is well. I do like the way this looks now with the use of the error monad!
</blockquote></details>

Done. A magic was learnt.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

mseri:
*[lib/network_utils.ml, line 291 at r2](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3XZkQLN2LiJunWR2UO:-L3gtgIdNp0KkNwcrf3b:b-hlbidb) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/f5d7fe7450dffc52bc65c25801df72f93d0598b1/lib/network_utils.ml#L291)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

SR-IOV actually ;)
</blockquote></details>

Then it needs changes in all the error lines :P

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

mseri:
*[lib/network_utils.ml, line 495 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3brMWwSdEUE0I_ZNEA:-L3gtqk54G4_xPuEz9LX:b-bpeenx) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L495)):*
<details><summary><i>Previously, xiewei20082008 (Wei Xie) wrote…</i></summary><blockquote>

For our currently supported NICs, all those NICs support MAC configuration and VLAN configuration, but some NICs do not support Rate configuration. We pass the `Fail_to_set_vf_rate` as a explicit error to XAPI to let XAPI know that this is not a general error but an error of hardware itself.
</blockquote></details>

Even if it's not yet used by xapi, I don't see the problem in making the errors better and more explicit when we know exactly what the errors are

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

xiewei20082008:
Review status: 1 of 4 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[lib/network_utils.ml, line 291 at r2](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3XZkQLN2LiJunWR2UO:-L3gbcwK1T6K7GzQyLFk:b-f3op7n) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/f5d7fe7450dffc52bc65c25801df72f93d0598b1/lib/network_utils.ml#L291)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

Then it needs changes in all the error lines :P
</blockquote></details>

No problem. I will fix them all tomorrow. ；）

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

xiewei20082008:
Review status: 1 of 4 files reviewed at latest revision, 16 unresolved discussions.

---

*[lib/network_utils.ml, line 276 at r2](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3X_9eXuZkS124uEfUC:-L3kzQ9nDT63v8sPAqlC:b-dbtpzf) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/f5d7fe7450dffc52bc65c25801df72f93d0598b1/lib/network_utils.ml#L276)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

And `write_one_line` from this module?
</blockquote></details>

Done. Tried this one and it works. 

---

*[lib/network_utils.ml, line 291 at r2](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3XZkQLN2LiJunWR2UO:-L3kze2Vz1RDPmkmhXVC:b-896fix) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/f5d7fe7450dffc52bc65c25801df72f93d0598b1/lib/network_utils.ml#L291)):*
<details><summary><i>Previously, xiewei20082008 (Wei Xie) wrote…</i></summary><blockquote>

No problem. I will fix them all tomorrow. ；）
</blockquote></details>

Done.

---

*[lib/network_utils.ml, line 495 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3brMWwSdEUE0I_ZNEA:-L3kzMRkBAQ_-djzQmXY:b-896fix) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L495)):*
<details><summary><i>Previously, mseri (Marcello Seri) wrote…</i></summary><blockquote>

Even if it's not yet used by xapi, I don't see the problem in making the errors better and more explicit when we know exactly what the errors are
</blockquote></details>

Done.

---

*[lib/network_utils.ml, line 1372 at r3](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3ZwDbbQGcc1tOEJT24:-L3kzO26S787k9ttHskc:b-896fix) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c9db895e9969bd2af8bc4fe2a6a196a7102a3ec0/lib/network_utils.ml#L1372)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

OK cool, it would be good to explain this in a comment in the code, otherwise someone might decide to "optimise" the code later.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

robhoes:
Reviewed 1 of 3 files at r4, 2 of 2 files at r6.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[networkd/network_server.ml, line 147 at r6](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3nGSeMnaXBOqyOEPjt:-L3nGSeMnaXBOqyOEPju:bcs7u6e) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/e48b4612b77ff05dff3c0c073851be2f94af4635/networkd/network_server.ml#L147)):*
> ```OCaml
> 	*)
> 	let parse_modprobe_conf_internal file_path driver option =
> 		(* Initially I did not use ref here, but afterward changed to ref to make the code easy to read *)
> ```

Please remove this comment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

xiewei20082008:
Review status: 3 of 4 files reviewed at latest revision, 4 unresolved discussions.

---

*[networkd/network_server.ml, line 147 at r6](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3nGSeMnaXBOqyOEPjt:-L3zi4Znkq7_ZDGGp0Oz:b-896fix) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/e48b4612b77ff05dff3c0c073851be2f94af4635/networkd/network_server.ml#L147)):*
<details><summary><i>Previously, robhoes (Rob Hoes) wrote…</i></summary><blockquote>

Please remove this comment.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

robhoes:
:lgtm:

---

Reviewed 1 of 1 files at r7.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-:-L40ZQG4WqWy7BTVbpp6:bnfp4nl)*

mseri:
*[networkd/network_server.ml, line 130 at r7](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L40_NF6UbC-EE37Uajw:-L40_NF6UbC-EE37Uajx:b-bs3ylg) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/c06b111342d615ca3a7b717a52f82965102f7dac/networkd/network_server.ml#L130)):*
> ```OCaml
> 		if Sysfs.get_sriov_maxvfs dev = 0 then [] else ["sriov"]
> 
> 	(* To enable SR-IOV via modprobe configuration, we add a line like `options igb max_vfs=7,7,7`		
> ```

This is not important, but can you fix this comment before merging? Some newlines went missing

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

lindig:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[lib/network_utils.ml, line 277 at r1](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126#-L3Dz6Y1o3qR_gmIAvqt:-L3Dz6Y1o3qR_gmIAvqu:b-7tozx2) ([raw file](https://github.com/xapi-project/xcp-networkd/blob/da1f0ea67d0aa0d0fef138fdd9c094cf6bb23501/lib/network_utils.ml#L277)):*
> ```OCaml
> 			|> String.trim
> 			|> int_of_string
> 			|> (-) 1 (* maxvfs is totalvfs -1, as totalvfs is PF num + VF num *)
> ```

This computes `1 - sriov_totalvfs`. I assume you want `sriovfs -1`. Use :100: 

`|> fun n -> n - 1`
 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/xapi-project/xcp-networkd/126)*

robhoes:
> This is not important, but can you fix this comment before merging? Some newlines went missing

Let's do this later :)

PR #132, opened 2018-03-07
CA-236855: VM.clean_shutdown task not completing on slave

YarsinCitrix:
From some historical test, it is found xapi (events_watch thread, for xenopsd events) is stuck some where some time. Further analysis found it is due to some earlier xapi->xcp-networkd call did not get response hence some mutex is not release so later use of the mutex is blocked. The reason the xapi->xcp-networkd is not responsed is further due to xcp-networkd to PVS proxy is not responded. After discussion with Rob and Jon, it is decided to add a time guard for xcp-networkd for the call to PVSproxy.

The implementation added a timebox call, in which another thread is created to do the actual RPC call, and the original thread will wait for some configurable timeout seconds, and proceed if the call did not ended.

### DIFF
--- a/networkd/network_server.ml
+++ b/networkd/network_server.ml
@@ -1016,20 +1016,31 @@ module PVS_proxy = struct
     
     let do_call call =
             let timebox ~timeout f =
-                let fd_in, fd_out = Unix.pipe () in
+				let fd_mutex = Xapi_stdext_threads.Threadext.Mutex.create () in
+				let fd_in, fd_out = Unix.pipe () in
+				let fd_closed = ref false in
+				let close_fd fd = Xapi_stdext_threads.Threadext.Mutex.execute fd_mutex
+					( fun () ->
+						if not !fd_closed then begin
+							fd_closed := true;
+							Unix.close fd
+						end
+					)
+				in
                 let _ =  Thread.create
                     (fun () ->
-                    f;
+                    f ();
                     debug "Complete execution in thread.";
-                    Unix.close fd_out) () in
+                    close_fd fd_out) () in
                 if Thread.wait_timed_read fd_in timeout then
                     debug "Timed execution completion."
                 else
                     debug "Timed execution timed out.";
-                Unix.close fd_in
+                Unix.close fd_in;
+                close_fd fd_out
             in
             
-            let rpc_call =
+            let rpc_call () =
                 try
                     Jsonrpc_client.with_rpc ~path:!path ~call ()
                 with e ->
--- a/networkd/network_server.ml
+++ b/networkd/network_server.ml
@@ -24,6 +24,7 @@ let network_conf = ref "/etc/xcp/network.conf"
 let config : config_t ref = ref Network_config.empty_config
 let backend_kind = ref Openvswitch
 let enic_workaround_until_version = ref "2.3.0.30"
+let pvsproxy_rpc_timeout = ref 300.0
 
 let legacy_management_interface_start () =
 	try
@@ -1012,13 +1013,31 @@ module PVS_proxy = struct
     open S.PVS_proxy
 
     let path = ref "/opt/citrix/pvsproxy/socket/pvsproxy"
+    
+    let do_call call =
+            let timebox ~timeout f =
+                let fd_in, fd_out = Unix.pipe () in
+                let _ =  Thread.create
+                    (fun () ->
+                    f;
+                    debug "Complete execution in thread.";
+                    Unix.close fd_out) () in
+                if Thread.wait_timed_read fd_in timeout then
+                    debug "Timed execution completion."
+                else
+                    debug "Timed execution timed out.";
+                Unix.close fd_in
+            in
+            
+            let rpc_call =
+                try
+                    Jsonrpc_client.with_rpc ~path:!path ~call ()
+                with e ->
+                    error "Error when calling PVS proxy: %s" (Printexc.to_string e);
+                    raise PVS_proxy_connection_error
+            in
 
-	let do_call call =
-		try
-			Jsonrpc_client.with_rpc ~path:!path ~call ()
-		with e ->
-			error "Error when calling PVS proxy: %s" (Printexc.to_string e);
-			raise PVS_proxy_connection_error
+            timebox ~timeout:!pvsproxy_rpc_timeout rpc_call
 
 	let configure_site dbg config =
 		debug "Configuring PVS proxy for site %s" config.site_uuid;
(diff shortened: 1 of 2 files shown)


networkd/network_server.ml
```diff
@@ -1012,13 +1013,31 @@ module PVS_proxy = struct
...
+                    debug "Complete execution in thread.";
+                    Unix.close fd_out) () in
+                if Thread.wait_timed_read fd_in timeout then
+                    debug "Timed execution completion."
+                else
+                    debug "Timed execution timed out.";
+                Unix.close fd_in
+            in
+            
+            let rpc_call =
```
mseri: I don't understand... from what this code looks like, `rpc_call` is the actual value of the call, it's not running in a separate thread. Also these are OCaml threads, they are not really parallel. Also it seems that nothing is writing in `fd_in`, so that would stay blocked forever. Finally, I don't see a `Thread.join` or some other mean to make sure we execute and complete the thread. I think this code is not doing what it should

mseri: Even if one adds the necessary `()` to run the call in the thread, I think this it is tricky to get right and the code above is not what we need. There is an infamous stackoverflow question that shows some options with different threadoffs and how tricky they are: https://stackoverflow.com/questions/35001959/interrupt-a-call-in-ocaml

Note in addition that one has to be very careful when dealing with `Thread.kill` and it generally not recommended.

edwintorok: Would be better to add the timeout on the socket itself, you can try `setsockopt` to set timeouts on individual reads, or use `select` to wait for available input or a timeout (although thats not so easy to do because the code deals with buffered channels, not file descriptors directly)

YarsinCitrix: @mseri , this timebox call is following the style of implementation from the same named function timebox in ocaml/xapi/helpers.ml, which seems to be already used in xenopsd. And I did test both sunny day and rainy day scenario, refer to the ca ticket log, it works as expected. The issue you mentioned in the post seems not apply to the current implementation.

YarsinCitrix: @edwintorok , Rob suggest me to do this on the call, I guess the reason is Jon is trying to add something at lower level for the RPC call, like on socket as you mentioned.

lindig: This is the `timebox` code from xapi. I doubt that the code above can work because the `Thread.create` does not apply function `f`. It is applied in the code below.

```
let timebox ~timeout ~otherwise f =
  let fd_in, fd_out = Unix.pipe () in
  let result = ref otherwise in
  let _ =  Thread.create
      (fun () ->
         (try
            let r = f () in
            result := fun () -> r
          with e ->
            result := fun () -> raise e);
         Unix.close fd_out) () in
  let _ = Thread.wait_timed_read fd_in timeout in
  Unix.close fd_in;
  !result ()
```

networkd/network_server.ml
```diff
@@ -1012,13 +1013,31 @@ module PVS_proxy = struct
...
+                try
+                    Jsonrpc_client.with_rpc ~path:!path ~call ()
```
robhoes: Removing the `()` at the end would be a good start. Otherwise (as @mseri said), this `rpc_call` won't be a function, the this would block here immediately before even creating any threads.

networkd/network_server.ml
robhoes: @edwintorok I'm not sure exactly how `Thread.wait_timed_read` (from the standard library) is implemented, but it may just use `select` internally. The docs say "Suspend the execution of the calling thread until at least one character is available for reading ". This xapi function confuses me as well, because nothing is written to `fd_in`.

robhoes: Ah, here it is:
```
let wait_timed_read fd d =
  match Unix.select [fd] [] [] d with ([], _, _) -> false | (_, _, _) -> true
```

networkd/network_server.ml
```diff
@@ -1012,13 +1013,31 @@ module PVS_proxy = struct
...
-	let do_call call =
-		try
-			Jsonrpc_client.with_rpc ~path:!path ~call ()
-		with e ->
-			error "Error when calling PVS proxy: %s" (Printexc.to_string e);
-			raise PVS_proxy_connection_error
+    
+    let do_call call =
+            let timebox ~timeout f =
+                let fd_in, fd_out = Unix.pipe () in
```
krizex: If the `f` hangs forever, the `fd_out` will leak, so we should close `fd_out` once timeouted.
`Helpers.timebox` has the same issue. https://github.com/xapi-project/xen-api/blob/master/ocaml/xapi/helpers.ml#L1133

networkd/network_server.ml
edwintorok: I think closing the other side of the pipe wakes up the select, because there is now you can successfully read 0 bytes without blocking, i.e. EOF.

edwintorok: https://github.com/ocaml/ocaml/pull/1656 lets improve the confusing documentation on `wait_timed_read`.

YarsinCitrix:
@mseri , @robhoes  and @krizex , folks, sorry I made a mistake there in previous code! I now have updated the code to reflect changes according to comments, a. making the rpc call a function passing to timebox and executing in the new thread; and b. enhance the fd closing logic with a flag guarded by mutex so no racing condition if both threads are working at the moment.

networkd/network_server.ml
```diff
@@ -1012,13 +1013,42 @@ module PVS_proxy = struct
...
+				let fd_closed = ref false in
+				let close_fd fd = Xapi_stdext_threads.Threadext.Mutex.execute fd_mutex
+					( fun () ->
+						if not !fd_closed then begin
+							fd_closed := true;
+							Unix.close fd
+						end
+					)
+				in
+                let _ =  Thread.create
```
krizex: indentation

networkd/network_server.ml
```diff
@@ -1012,13 +1013,42 @@ module PVS_proxy = struct
...
-	let do_call call =
-		try
-			Jsonrpc_client.with_rpc ~path:!path ~call ()
-		with e ->
-			error "Error when calling PVS proxy: %s" (Printexc.to_string e);
-			raise PVS_proxy_connection_error
+    
+    let do_call call =
+            let timebox ~timeout f =
+				let fd_mutex = Xapi_stdext_threads.Threadext.Mutex.create () in
```
krizex: Is mutex amount limited in OCaml? If yes, we should move this mutex to the module level to make it a module member to avoid mutex leak.
I did some google search, the result shows the implementation in libc does not restrict the amount of mutex.

edwintorok: It is better to have more fine grained locks, creating a new mutex should be as cheap as allocating some memory for it, the GC will take care of claiming it back.

YarsinCitrix:
@mseri and @robhoes, any further comments based on the current version?

mseri:
I am closing this. Further discussion is going to be on the ticket

PR #133, opened 2018-03-07
Port sriov to ppx

minli1:
This is for after rebase master then changes the interface to PPX.

### DIFF
--- a/networkd/network_server.ml
+++ b/networkd/network_server.ml
@@ -18,6 +18,7 @@ open Xapi_stdext_std
 open Xapi_stdext_unix
 open Xapi_stdext_monadic
 
+module S = Network_interface.Interface_API(Idl.GenServerExn ())
 module D = Debug.Make(struct let name = "network_server" end)
 open D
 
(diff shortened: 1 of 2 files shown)


mseri:
Something is wrong. Why travis fails with:
```ocaml
File "lib/network_config.ml", line 109, characters 49-64:
Error: Unbound value typ_of_config_t
Hint: Did you mean rpc_of_config_t?
```
This seems rebased on top of the ppx'd networkd, is there some piece that got missing in the process?

networkd/network_server.ml
```diff
@@ -122,7 +123,9 @@ let need_enic_workaround () =
 		| None -> false )
 
 module Sriov = struct
```
mseri: This module should be moved down after the PVS_proxy module, there should be already an `S` module defined there: https://github.com/xapi-project/xcp-networkd/blob/master/networkd/network_server.ml#L1009

robhoes: ...or remove the definition of `S` below in favour of the one above. I would suggest to not move the module in this PR, because it causes unnecessary conflicts in another PR, and this PR isn't actually introducing the module.

mseri: We looked into this and it looks like the modules S was moved in different commits, and the change went lost in the rebase.

mseri: Also, the module needs to be where it is, because it is used in `get_capabilities`: 
https://github.com/minli1/xcp-networkd/blob/3a9e6d460deafdbce66a211cbf479f1723673666/networkd/network_server.ml#L552

Again, this is an older change that got hidden by the rebase

PR #138, opened 2018-04-09
CA-287340: Slave reboot with unexpect enabled SRIOV.

xiewei20082008:
Even if the current numvfs equals to the target numvfs, it is still needed to update SR-IOV modprobe config file, as the SR-IOV enabing takes effect after reboot. For example, a user enables SR-IOV and disables it immediately without a reboot.

Signed-off-by: Wei Xie <wei.xie@citrix.com>

### DIFF
--- a/networkd/network_server.ml
+++ b/networkd/network_server.ml
@@ -148,11 +148,15 @@ let get_capabilities dev =
 		| Some vf_param ->
 			debug "enable SR-IOV on a device: %s via modprobe" dev;
 			(if enable then Modprobe.get_maxvfs driver config else Ok 0) >>= fun numvfs ->
-			if numvfs = Sysfs.get_sriov_numvfs dev then Ok Modprobe_successful
-			else begin
-				Modprobe.config_sriov driver vf_param numvfs >>= fun _ ->
+			(* CA-287340: Even if the current numvfs equals to the target numvfs,
+			it is still needed to update SR-IOV modprobe config file, as the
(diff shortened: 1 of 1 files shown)


networkd/network_server.ml
```diff
@@ -148,7 +148,13 @@ let get_capabilities dev =
 		| Some vf_param ->
 			debug "enable SR-IOV on a device: %s via modprobe" dev;
 			(if enable then Modprobe.get_maxvfs driver config else Ok 0) >>= fun numvfs ->
-			if numvfs = Sysfs.get_sriov_numvfs dev then Ok Modprobe_successful
+			if numvfs = Sysfs.get_sriov_numvfs dev then
```
mseri: So we could actually simplify the code a bit further and do 
```ocaml
[...] >>= fun numvfs ->
(* CA-287340: Even if the current numvfs equals to the target numvfs,
it is still needed to update SR-IOV modprobe config file, as the
SR-IOV enabing takes effect after reboot. For example, a user
enables SR-IOV and disables it immediately without a reboot.*)
Modprobe.config_sriov driver vf_param numvfs >>= fun _ ->
if numvfs = Sysfs.get_sriov_numvfs dev then
    Ok Modprobe_successful
else
    Ok Modprobe_successful_requires_reboot
```

Overall seems fine

xiewei20082008: Done. Thank you :)

PR #140, opened 2018-05-15
CA-236855: VM.clean_shutdown task not completing on slave

YarsinCitrix:
* CA-236855: VM.clean_shutdown task not completing on slave

According to the discussion (also recorded in the ticket), this change
will use the newly created read_timeout and write_timeout to process
input and output from outside. Some updates according to Nanjing peer review.

Modified test case to remove the cases that is not applied.

### DIFF
--- a/lib/jsonrpc_client.ml
+++ b/lib/jsonrpc_client.ml
@@ -32,26 +32,29 @@ let timeout_read fd timeout =
 	let read_start = Mtime_clock.counter () in
 	let get_total_used_time () = Mtime.Span.to_uint64_ns (Mtime_clock.count read_start) in
 	let rec inner max_time max_bytes =
-		let (ready_to_read, _, _) = Unix.select [fd] [] [] (to_s max_time) in
+		let (ready_to_read, _, _) = try Unix.select [fd] [] [] (to_s max_time) with
+			(* in case the unix.select call fails in situation like interrupt *)
+			| _ -> [], [], []
+		in
 		(* This is not accurate the calculate time just for the select part. However, we
 		 * think the read time will be minor comparing to the scale of tens of seconds.
 		 * the current style will be much concise in code. *)
-		let used_time = get_total_used_time () in
-		let remain_time = Int64.sub timeout used_time in
+		let remain_time = 
+			let used_time = get_total_used_time () in
+			Int64.sub timeout used_time
+		in
 		if remain_time < 0L then 
 		begin
 			debug "Timeout after read %d" (Buffer.length buf);
 			raise Timeout
 		end;
-		if List.mem fd ready_to_read
-		then
+		if List.mem fd ready_to_read then
 		begin
 			let bytes = Bytes.make 4096 '\000' in
 			match Unix.read fd bytes 0 4096 with
 				| 0 -> Buffer.contents buf (* EOF *)
 				| n ->
-					if n > max_bytes
-					then
+					if n > max_bytes then
 					begin
 						debug "exceeding maximum read limit %d, clear buffer" !json_rpc_max_len;
 						Buffer.clear buf;
@@ -75,30 +78,35 @@ let timeout_read fd timeout =
 let timeout_write filedesc total_length data response_time =
 	let write_start = Mtime_clock.counter () in
 	let get_total_used_time () = Mtime.Span.to_uint64_ns (Mtime_clock.count write_start) in
-	let rec inner_write filedesc offset length remain_time =
-		let (_, ready_to_write, _) = Unix.select [] [filedesc] [] (to_s remain_time) in 
-		let used_time = get_total_used_time () in
-		let new_remain_time = Int64.sub response_time used_time in
-		if new_remain_time < 0L then
+	let rec inner_write offset max_time =
+		let (_, ready_to_write, _) = try Unix.select [] [filedesc] [] (to_s max_time) with
+			(* in case the unix.select call fails in situation like interrupt *)
+			| _ -> [], [], []
+		in
+		let remain_time = 
+			let used_time = get_total_used_time () in
+			Int64.sub response_time used_time
+		in
+		if remain_time < 0L then
 		begin
 			debug "Timeout to write %d at offset %d" total_length offset;
 			raise Timeout
 		end;
 		if List.mem filedesc ready_to_write then 
 		begin
+			let length = total_length - offset in
 			let bytes_written = 
 				(try Unix.single_write filedesc data offset length with 
 				| Unix.Unix_error(Unix.EAGAIN,_,_)
 				| Unix.Unix_error(Unix.EWOULDBLOCK,_,_) -> 0)
 			in
 			let new_offset = offset + bytes_written in
-			let new_length = length - bytes_written in
-			if new_length = 0 then ()
-			else inner_write filedesc new_offset new_length new_remain_time
+			if length = bytes_written then ()
+			else inner_write new_offset remain_time
 		end
-		else inner_write filedesc offset length new_remain_time
+		else inner_write offset remain_time
 	in
-	inner_write filedesc 0 total_length response_time
+	inner_write 0 response_time
 
 let with_rpc ?(version=Jsonrpc.V2) ~path ~call () =
 	let uri = Uri.of_string (Printf.sprintf "file://%s" path) in
--- a/lib/jsonrpc_client.ml
+++ b/lib/jsonrpc_client.ml
@@ -21,8 +21,8 @@ exception Timeout
 exception Read_error
 
 let json_rpc_max_len = ref 65536 (* Arbitrary maximum length of RPC response *)
-let json_rpc_read_timeout = ref 60000000000L
-let json_rpc_write_timeout = ref 60000000000L
+let json_rpc_read_timeout = ref 60000000000L (* timeout value in ns when reading RPC response *)
+let json_rpc_write_timeout = ref 60000000000L (* timeout value in ns when writing RPC request *)
 
 let to_s  s = (Int64.to_float s) *. 1e-9
 
@@ -34,7 +34,7 @@ let timeout_read fd timeout =
 	let rec inner max_time max_bytes =
 		let (ready_to_read, _, _) = try Unix.select [fd] [] [] (to_s max_time) with
 			(* in case the unix.select call fails in situation like interrupt *)
-			| _ -> [], [], []
+			| Unix.Unix_error(Unix.EINTR,_,_) -> [], [], []
 		in
 		(* This is not accurate the calculate time just for the select part. However, we
 		 * think the read time will be minor comparing to the scale of tens of seconds.
@@ -81,7 +81,7 @@ let timeout_write filedesc total_length data response_time =
 	let rec inner_write offset max_time =
 		let (_, ready_to_write, _) = try Unix.select [] [filedesc] [] (to_s max_time) with
 			(* in case the unix.select call fails in situation like interrupt *)
-			| _ -> [], [], []
+			| Unix.Unix_error(Unix.EINTR,_,_) -> [], [], []
 		in
 		let remain_time = 
 			let used_time = get_total_used_time () in
--- a/lib/jsonrpc_client.ml
+++ b/lib/jsonrpc_client.ml
@@ -17,49 +17,95 @@
 module D = Debug.Make(struct let name = "jsonrpc_client" end)
 open D
 
-let input_json_object fin =
-	let buf = Buffer.create 1024 in
-	let brace_cnt = ref 0 in
-	let in_string = ref false in
-	let last_char () = Buffer.nth buf (Buffer.length buf - 1) in
-	let rec get () =
-		let c = input_char fin in
+exception Timeout
+exception Read_error
+
+let json_rpc_max_len = ref 65536 (* Arbitrary maximum length of RPC response *)
+let json_rpc_read_timeout = ref "60000000000"
+let json_rpc_write_timeout = ref "60000000000"
+
+let to_s  s = (Int64.to_float s) *. 1e-9
+
+(* Read the entire contents of the fd, of unknown length *)
+let timeout_read fd timeout =
+	let buf = Buffer.create !json_rpc_max_len in
+	let read_start = Mtime_clock.counter () in
+	let get_total_used_time () = Mtime.Span.to_uint64_ns (Mtime_clock.count read_start) in
+	let rec inner max_time max_bytes =
+		let (ready_to_read, _, _) = Unix.select [fd] [] [] (to_s max_time) in
+		(* This is not accurate the calculate time just for the select part. However, we
+		 * think the read time will be minor comparing to the scale of tens of seconds.
+		 * the current style will be much concise in code. *)
+		let used_time = get_total_used_time () in
+		let remain_time = Int64.sub timeout used_time in
+		if remain_time < 0L then 
 		begin
-			match c with
-			| '{' when not !in_string -> brace_cnt := !brace_cnt + 1
-			| '}' when not !in_string -> brace_cnt := !brace_cnt - 1
-			| '"' when !in_string && (last_char () <> '\\') -> in_string := false
-			| '"' when not !in_string -> in_string := true
-			| _ -> ()
+			debug "Timeout after read %d" (Buffer.length buf);
+			raise Timeout
 		end;
-		Buffer.add_char buf c;
-		if !brace_cnt > 0 then
-			get ()
+		if List.mem fd ready_to_read
+		then
+		begin
+			let bytes = Bytes.make 4096 '\000' in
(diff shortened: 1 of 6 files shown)


lib/jsonrpc_client.ml
```diff
@@ -17,49 +17,95 @@
...
-			| '"' when !in_string && (last_char () <> '\\') -> in_string := false
-			| '"' when not !in_string -> in_string := true
-			| _ -> ()
+			debug "Timeout after read %d" (Buffer.length buf);
+			raise Timeout
 		end;
-		Buffer.add_char buf c;
-		if !brace_cnt > 0 then
-			get ()
+		if List.mem fd ready_to_read
```
krizex: It's better to put `if` and `then` in the same line to reduce the length of the code block

lib/jsonrpc_client.ml
```diff
@@ -17,49 +17,95 @@
...
+					end
+					else
+					begin
+						Buffer.add_subbytes buf bytes 0 n;
+						inner remain_time (max_bytes - n)
+					end
+				| exception Unix.Unix_error(err,_,_) when err = Unix.EAGAIN || err = Unix.EWOULDBLOCK ->
+					inner remain_time max_bytes
+		end
+		else inner remain_time max_bytes
```
krizex: We are already timeout here, why not raise `Timout`?

lib/jsonrpc_client.ml
```diff
@@ -17,49 +17,95 @@
...
+	let buf = Buffer.create !json_rpc_max_len in
+	let read_start = Mtime_clock.counter () in
+	let get_total_used_time () = Mtime.Span.to_uint64_ns (Mtime_clock.count read_start) in
+	let rec inner max_time max_bytes =
+		let (ready_to_read, _, _) = Unix.select [fd] [] [] (to_s max_time) in
+		(* This is not accurate the calculate time just for the select part. However, we
+		 * think the read time will be minor comparing to the scale of tens of seconds.
+		 * the current style will be much concise in code. *)
+		let used_time = get_total_used_time () in
+		let remain_time = Int64.sub timeout used_time in
```
krizex: I prefer to hide the variable which won't be referred later. 
```
let remain_time = 
  let used_time = get_total_used_time () in
  Int64.sub timeout used_time in
```

lib/jsonrpc_client.ml
```diff
@@ -17,49 +17,95 @@
...
-	result
+(* Write as many bytes to a file descriptor as possible from data before a given clock time. *)
+(* Raises Timeout exception if the number of bytes written is less than the specified length. *)
+(* Writes into the file descriptor at the current cursor position. *)
+let timeout_write filedesc total_length data response_time =
+	let write_start = Mtime_clock.counter () in
+	let get_total_used_time () = Mtime.Span.to_uint64_ns (Mtime_clock.count write_start) in
+	let rec inner_write filedesc offset length remain_time =
+		let (_, ready_to_write, _) = Unix.select [] [filedesc] [] (to_s remain_time) in 
+		let used_time = get_total_used_time () in
```
krizex: Same with the comment of `used_time` in `timeout_read`.

lib/jsonrpc_client.ml
```diff
@@ -17,49 +17,95 @@
...
+let json_rpc_write_timeout = ref "60000000000"
+
+let to_s  s = (Int64.to_float s) *. 1e-9
+
+(* Read the entire contents of the fd, of unknown length *)
+let timeout_read fd timeout =
+	let buf = Buffer.create !json_rpc_max_len in
+	let read_start = Mtime_clock.counter () in
+	let get_total_used_time () = Mtime.Span.to_uint64_ns (Mtime_clock.count read_start) in
+	let rec inner max_time max_bytes =
```
krizex: How about renaing `max_time` to `remain_time` to keep consistent with the `inner_write` function of `timeout_write` below?

lib/jsonrpc_client.ml
```diff
@@ -17,49 +17,95 @@
...
-	close_in fin;
-	debug "Shut down.";
-	result
+(* Write as many bytes to a file descriptor as possible from data before a given clock time. *)
+(* Raises Timeout exception if the number of bytes written is less than the specified length. *)
+(* Writes into the file descriptor at the current cursor position. *)
+let timeout_write filedesc total_length data response_time =
+	let write_start = Mtime_clock.counter () in
+	let get_total_used_time () = Mtime.Span.to_uint64_ns (Mtime_clock.count write_start) in
+	let rec inner_write filedesc offset length remain_time =
```
krizex: We could make this inner function as a closure so we can reduce its parameter to
```
let rec inner_write offset remain_time =
```
So that `filedesc` will use the same parameter of `timeout_write` and the `length` could be calculated with `total_length - offset` once we need it.

lib/jsonrpc_client.ml
```diff
@@ -17,49 +17,95 @@
...
-		receive fin
-	)
-
+	let uri = Uri.of_string (Printf.sprintf "file://%s" path) in
+	Open_uri.with_open_uri uri (fun s ->
+		Unix.set_nonblock s;
+		let req = Bytes.of_string (Jsonrpc.string_of_call ~version call) in
+		timeout_write s (Bytes.length req) req (Int64.of_string !json_rpc_write_timeout);
+		let res = timeout_read s (Int64.of_string !json_rpc_read_timeout) in
+		debug "Response: %s" res;
```
krizex: If print the response, it is better to print the request at same time.

lib/jsonrpc_client.ml
```diff
@@ -17,49 +17,95 @@
...
-	let brace_cnt = ref 0 in
-	let in_string = ref false in
-	let last_char () = Buffer.nth buf (Buffer.length buf - 1) in
-	let rec get () =
-		let c = input_char fin in
+exception Timeout
+exception Read_error
+
+let json_rpc_max_len = ref 65536 (* Arbitrary maximum length of RPC response *)
+let json_rpc_read_timeout = ref "60000000000"
```
krizex: I prefer to make the configuration unit with **ms** rather than **ns** to make it user friendly.
It's really difficult for me to count the `0` and I think **ms** is precise enough for timeout.

lib/jsonrpc_client.ml
mseri: Are we sure we want to print either of them? How often do we call this function?

lib/jsonrpc_client.ml
mseri: Why? Couldn't it be that something else triggered the select?

krizex: We are only watching one fd, will the `select` return when other fds which we are not watching is ready to read?

mseri: Yes, you are right. We could do `if remain_time < 0L or not (List.mem fd ready_to_read) then ...` and get rid of the second `if` altogether.

krizex: Exactly, but do we need to consider the third element of the return value of `select` which contains the fds that exception occurs?

mseri: Does it return something even if we are not watching it for exceptional cases?

krizex: I don't think so, so we don't need to care about that.

YarsinCitrix: @krizex , I cannot agree here. It is how and when select will return is not accurately defined by Ocaml lib, so I think the current code is necessary. Usually in c coding with native OS select, it is necessary to calculate the real time after returned, as select could prematurely return as interrupted by other signal or events from the IO. I doubt Ocaml lib will wrap something to make the unix.select more 'robust' to such event, as usually such event is practically necessary in some situation.

minishrink:
One suggestion: running this through ocp-indent, as you've got some large indentations in there.

lib/jsonrpc_client.ml
robhoes: And why not make it `int64`s right away? Then you don't need the `Int64.of_string` for each call below.

lib/jsonrpc_client.ml
```diff
@@ -17,49 +17,95 @@
...
-	debug "Shut down.";
-	result
+(* Write as many bytes to a file descriptor as possible from data before a given clock time. *)
+(* Raises Timeout exception if the number of bytes written is less than the specified length. *)
+(* Writes into the file descriptor at the current cursor position. *)
+let timeout_write filedesc total_length data response_time =
+	let write_start = Mtime_clock.counter () in
+	let get_total_used_time () = Mtime.Span.to_uint64_ns (Mtime_clock.count write_start) in
+	let rec inner_write filedesc offset length remain_time =
+		let (_, ready_to_write, _) = Unix.select [] [filedesc] [] (to_s remain_time) in
```
robhoes: As above, this may raise a `Unix_error` with `Unix.EINTR`, which you should catch.

lib/jsonrpc_client.ml
robhoes: So you should catch the `Unix_error` with argument `EINTR` from `select` with an exception handler. And `Unix.read` and `Unix.single_write` as well just in case. Remember the discussion about `really_write` from a while ago?

lib/jsonrpc_client.ml
```diff
@@ -17,49 +17,95 @@
...
+						debug "exceeding maximum read limit %d, clear buffer" !json_rpc_max_len;
+						Buffer.clear buf;
+						raise Read_error
+					end
+					else
+					begin
+						Buffer.add_subbytes buf bytes 0 n;
+						inner remain_time (max_bytes - n)
+					end
+				| exception Unix.Unix_error(err,_,_) when err = Unix.EAGAIN || err = Unix.EWOULDBLOCK ->
```
robhoes: And `exception End_of_file`?

lib/jsonrpc_client.ml
krizex: I think Rob is right, refer to https://github.com/ocaml/ocaml/blob/13945a71ed2616df9a4672cfc83f8f450ec437a8/otherlibs/unix/select.c#L96

lib/jsonrpc_client.ml
```diff
@@ -17,49 +17,103 @@
...
```
krizex: `| exception Unix.Unix_error(Unix.EAGAIN | Unix.EWOULDBLOCK,_,_)`  -> would be better

lib/jsonrpc_client.ml
```diff
@@ -17,49 +17,103 @@
...
+			debug "Timeout to write %d at offset %d" total_length offset;
+			raise Timeout
+		end;
+		if List.mem filedesc ready_to_write then 
+		begin
+			let length = total_length - offset in
+			let bytes_written = 
+				(try Unix.single_write filedesc data offset length with 
+				| Unix.Unix_error(Unix.EAGAIN,_,_)
+				| Unix.Unix_error(Unix.EWOULDBLOCK,_,_) -> 0)
```
krizex: Combine the two lines into
```
Unix.Unix_error(Unix.EAGAIN | Unix.EWOULDBLOCK,_,_) -> 0
```

lib/jsonrpc_client.ml
krizex: Why not define it as `int64`?

networkd/networkd.ml
```diff
@@ -64,6 +64,9 @@ let options = [
 	"dracut-cmd-path", Arg.Set_string Network_utils.dracut, (fun () -> !Network_utils.dracut), "Path to the Unix command dracut";
 	"dracut-timeout", Arg.Set_float Network_utils.dracut_timeout, (fun () -> string_of_float !Network_utils.dracut_timeout), "Default value for the dracut command timeout";
 	"modinfo-cmd-path", Arg.Set_string Network_utils.modinfo, (fun () -> !Network_utils.modinfo), "Path to the Unix command modinfo";
+	"json-rpc-max-len", Arg.Set_int Jsonrpc_client.json_rpc_max_len, (fun () -> string_of_int !Jsonrpc_client.json_rpc_max_len), "Maximum buffer size for Json RPC response";
+	"json-rpc-read-timeout", Arg.Set_string Jsonrpc_client.json_rpc_read_timeout, (fun () -> !Jsonrpc_client.json_rpc_read_timeout), "JSON RPC response read timeout value in ns";
```
krizex: What you get from user is always `string`, here you can convert the string value to `int64`

YarsinCitrix: Arg does not support int64. And I am not sure the impact vs risks introducing expansion to support int64 here.

krizex: Once the configuration unit changes to `ms`, int is large enough to store the value. Just convert from int to int64 while using, it costs less than convert from string to int64.

YarsinCitrix: the mtime/mtime_clock uses int64, if we do int/ms, then that means extra conversion. Why are you obsessed with just a fewer zeros?

mseri: I think the best reason to avoid the manual conversion from string, is that if we set the argument as an int we will fail to run immediately when we load the variables. 

I don't think converting int to int64 or int64 from string later will be such a burden, we are doing this in a IO bound function, the IO will take the most of the time anyways.

robhoes: If the ref were of type `int64`, then I think you can simply replace

`Arg.Set_int Jsonrpc_client.json_rpc_max_len`

by

`Arg.Int (fun x -> Jsonrpc_client.json_rpc_max_len := Int64.of_int x)` to make it work.

lib/jsonrpc_client.ml
robhoes: `EINTR`?

lib/jsonrpc_client.ml
robhoes: `EINTR`?

networkd/networkd.ml
```diff
@@ -64,6 +64,9 @@ let options = [
 	"dracut-cmd-path", Arg.Set_string Network_utils.dracut, (fun () -> !Network_utils.dracut), "Path to the Unix command dracut";
 	"dracut-timeout", Arg.Set_float Network_utils.dracut_timeout, (fun () -> string_of_float !Network_utils.dracut_timeout), "Default value for the dracut command timeout";
 	"modinfo-cmd-path", Arg.Set_string Network_utils.modinfo, (fun () -> !Network_utils.modinfo), "Path to the Unix command modinfo";
+	"json-rpc-max-len", Arg.Set_int Jsonrpc_client.json_rpc_max_len, (fun () -> string_of_int !Jsonrpc_client.json_rpc_max_len), "Maximum buffer size for Json RPC response";
+	"json-rpc-read-timeout", Arg.Int (fun x -> Jsonrpc_client.json_rpc_read_timeout := Int64.(mul 1000L (of_int x))), (fun () -> Int64.(to_string (div !Jsonrpc_client.json_rpc_read_timeout 1000L))), "JSON RPC response read timeout value in ms";
```
krizex: Convert from `ms` to `ns`, should multiply by 1000000L.

krizex:
Looks good to me!
Please update the title of commit https://github.com/xapi-project/xcp-networkd/pull/140/commits/27517bbe7bc49b59d7496a4cefc22fc4943323d8

lib/jsonrpc_client.ml
```diff
@@ -17,49 +17,102 @@
...
+(* Write as many bytes to a file descriptor as possible from data before a given clock time. *)
+(* Raises Timeout exception if the number of bytes written is less than the specified length. *)
+(* Writes into the file descriptor at the current cursor position. *)
+let timeout_write filedesc total_length data response_time =
+	let write_start = Mtime_clock.counter () in
+	let get_total_used_time () = Mtime.Span.to_uint64_ns (Mtime_clock.count write_start) in
+	let rec inner_write offset max_time =
+		let (_, ready_to_write, _) = try Unix.select [] [filedesc] [] (to_s max_time) with
+			(* in case the unix.select call fails in situation like interrupt *)
+			| _ -> [], [], []
```
robhoes: I'm worried about the `_` here and think we should just catch `EINTR`, because that is an error that it is worth immediately retrying on. If an other, more persistent, error occurs, then we'll end up spinning in this loop until the timeout occurs. It's better to let the exception break the loop in that case.

lib/jsonrpc_client.ml
```diff
@@ -17,49 +17,102 @@
...
-	let brace_cnt = ref 0 in
-	let in_string = ref false in
-	let last_char () = Buffer.nth buf (Buffer.length buf - 1) in
-	let rec get () =
-		let c = input_char fin in
+exception Timeout
+exception Read_error
+
+let json_rpc_max_len = ref 65536 (* Arbitrary maximum length of RPC response *)
+let json_rpc_read_timeout = ref 60000000000L
```
robhoes: Let's add a comment with the units here to avoid confusion.

robhoes:
I've done a squash+merge.

PR #142, opened 2018-06-14
CA-291197: Toolstack no longer ignores junk at the end of JSON-RPC re…

YarsinCitrix:
…sponses from pvsproxy

This is resulted by CA-236855, in that the read strategy is changed to assume all good
data will come from PVSproxy side. However, in testing, it is found a bug in PVSproxy,
which was not fatal locally, but resulted issue in overall scenario. It is decided to
enhance the solution a little bit based on previous change.

The updates as to integrate some part in the old byte by byte read, where a function
is used to prescreen the data read from PVSproxy, and tailing junk is discarded, so
the only json part is further processed. Here we most copied the original function,
but added:
 1. to get input chars from a string;
 2. to tolerant heading spaces/junk (in screening);

### DIFF
--- a/lib/jsonrpc_client.ml
+++ b/lib/jsonrpc_client.ml
@@ -26,6 +26,40 @@ let json_rpc_write_timeout = ref 60000000000L (* timeout value in ns when writin
 
 let to_s  s = (Int64.to_float s) *. 1e-9
 
+let retrieve_json_str str =
+	let buf = Buffer.create 1024 in
+	let fetch_indx = ref 0 in
+	let str_len = String.length str in
+	let input_char () =
+		if !fetch_indx < str_len then
+		begin
+			let chr = String.get str !fetch_indx in
+			fetch_indx := !fetch_indx + 1;
+			chr
+		end
+		else raise Yojson.End_of_input
+	in
+	let brace_cnt = ref (-1) in (* made to tolerant heading space *)
+	let in_string = ref false in
+	let last_char () = Buffer.nth buf (Buffer.length buf - 1) in
+	let rec get () =
+		let c = input_char () in
+		begin
+			match c with
+			| '{' when not !in_string ->
+				if !brace_cnt = -1 then brace_cnt := 1
+				else brace_cnt := !brace_cnt + 1
+			| '}' when not !in_string -> brace_cnt := !brace_cnt - 1
+			| '"' when !in_string && (last_char () <> '\\') -> in_string := false
+			| '"' when not !in_string -> in_string := true
+			| _ -> ()
+		end;
+		Buffer.add_char buf c;
+		if !brace_cnt <> 0 then get ()
+	in
+	get ();
+	Buffer.contents buf
+
 (* Read the entire contents of the fd, of unknown length *)
 let timeout_read fd timeout =
 	let buf = Buffer.create !json_rpc_max_len in
@@ -113,6 +147,7 @@ let with_rpc ?(version=Jsonrpc.V2) ~path ~call () =
 		Unix.set_nonblock s;
 		let req = Bytes.of_string (Jsonrpc.string_of_call ~version call) in
 		timeout_write s (Bytes.length req) req !json_rpc_write_timeout;
-		let res = timeout_read s !json_rpc_read_timeout in
-		debug "Response: %s" res;
+		let buff = timeout_read s !json_rpc_read_timeout in
+		debug "Response: %s" buff;
+		let res = retrieve_json_str buff in
 		Jsonrpc.response_of_string res)
--- a/lib/jsonrpc_client.mli
+++ b/lib/jsonrpc_client.mli
@@ -19,6 +19,8 @@ val json_rpc_max_len : int ref
 val json_rpc_read_timeout : int64 ref
 val json_rpc_write_timeout : int64 ref
 
+val retrieve_json_str : string -> string
+
 val timeout_read : Unix.file_descr -> int64 -> string
 (** Do an JSON-RPC call to a server that is listening on a Unix domain 
  *  socket at the given path. *)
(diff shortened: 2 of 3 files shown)


YarsinCitrix:
I was trying to alter a little bit Yojson to get just one json out and ignores tailing junk, and then pass it to jsonrpc.
The Yojson part could be done, as most of the related functions are exposed in read.mli. However, functions like json_to_rpc in jsonrpc is not exposed, so we cannot create an updated jsonrpc module to adopt the json out of yojson from the input string. The other way left is just use the updated Yojson to get the json, from the initial output string, and then convert the json back to string, and pass it to jsonrpc as normal. This is considered a little bit complex and less efficient, after discussion with Yang. So we are trying to start with integrate the original code.

lindig:
Just some general remarks:

* I think a better name for `retrieve_json` would be `strip` or `strip_json`
* JSON is UTF8 encoded. Can we be sure that the manual parsing of JSON implemented in `retrieve_json` does not trip over that? I suspect we can because bytes in UTF8 that belong to a sequence have their highest bit set and thus can't contain characters like `"` or `{` that the scanner is looking for.
* I still would prefer a solution that does not try to implement a JSON parser and make both sides exchange proper JSON that does not require special massaging.

lib/jsonrpc_client.ml
```diff
@@ -26,6 +26,40 @@ let json_rpc_write_timeout = ref 60000000000L (* timeout value in ns when writin
...
+		else raise Yojson.End_of_input
+	in
+	let brace_cnt = ref (-1) in (* made to tolerant heading space *)
+	let in_string = ref false in
+	let last_char () = Buffer.nth buf (Buffer.length buf - 1) in
+	let rec get () =
+		let c = input_char () in
+		begin
+			match c with
+			| '{' when not !in_string ->
```
robhoes: I think it's a pity if we need to do all this low-level parsing here again (we just deleted a function like this). A JSON library should be doing it. Can we instead modify ocaml-rpc to have an option to filter out "junk" at the end of a JSON-RPC request?

edwintorok: Instead of implementing our own JSON parser wouldn't `Yojson.Basic.from_lexbuf ~stream:true` give us enough control to skip everything after the JSON value? (alternatively Jsonm should also give us similar low-level control)

edwintorok:
I agree with @lindig last point above, this sounds like a bug that should be fixed in pvsproxy and not here.

lib/jsonrpc_client.ml
YarsinCitrix: I tried, but ocaml-rpc does not expose many of its internals so we cannot simply rewrite the logic there. For Yojson, I made changes like following:
module Yojson_private = struct
  include Yojson.Safe

  let from_lexbuf v ?(stream = false) lexbuf =
      read_space v lexbuf;
      if read_eof lexbuf then
        raise Yojson.End_of_input
      else
        read_json v lexbuf

  let from_string ?buf ?fname ?lnum s =
    try
      let lexbuf = Lexing.from_string s in
      let v = init_lexer ?buf ?fname ?lnum () in
      from_lexbuf v lexbuf
    with End_of_input ->
      Yojson.json_error "Blank input data"
end

It now can just read the first json out and ignore what's left. However, as I stated above, this json cannot be sent back to jsonrpc. It does not accept json to convert to Rpc, from a module perspective. If anyone can suggest any simple way we can override jsonrpc's some internal logic, then I can try.

robhoes:
It is also being fixed in pvsproxy, but we should be more robust in xcp-networkd as well. And we were robust in this way until a recent patch, which is another reason to revert to this behaviour for compatibility.

lib/jsonrpc_client.ml
YarsinCitrix: Specifically, if we have functions like json_to_rpc and get_response exposed in interface for jsonrpc, then it could be a more straight forwarding change.

edwintorok: Could you add a `?(stream=false)` to from_string and propagata it to `from_lexbuf`?

YarsinCitrix: It is similar, could be done. But the point is really not how this get done in Yojson. It is we cannot get the json out of Yojson back to jsonrpc.

jonludlam:
we kind-of _have_ to fix this in xcp-networkd _as well_ as pvsproxy because pvsproxy won't be upgraded at the same time as xcp-networkd (pvsproxy is delivered separately). Ideally I'd like to remove this in the next release.

lib/jsonrpc_client.ml
jonludlam: I think a PR to ocaml-rpc to add a way to do streaming seems the best solution here.

YarsinCitrix: Shall we merge this in first, as there is a broken product waiting to be fixed/have work around?
Then we can take time to enhance/improve Yojson or ocaml-rpc and integrate those changes to refine codes here.

Besides, not sure what exactly you mean by streaming. If you mean to have ocaml-rpc to be able to get some json one by one, we may need to also largely change Yojson, coz the current Yojson only deal with complete and correct jsons even in streaming mode. And that also takes time. If you mean just get one rpc/json from the input (and ignores what's left) in ocaml-rpc, will they owner of ocaml-rpc accept such change as it seems incomplete?

YarsinCitrix: Anyway, the current change won't be worse than what's before CA-236855, where similar code to parse json works.

robhoes:
Meanwhile, the pvsproxy has been fixed to not have junk at the end of the message anymore, so the problem is now less urgent. However, we still want to fix it here as well (as hinted at above), but there is no need to make a temporary fix first. @jonludlam says that an update to ocaml-rpc to address this would not be too hard to do, and recommends doing that instead and closing this PR. So let's do that.

PR #143, opened 2018-06-21
CA-291197: Toolstack no longer ignores junk at the end of JSON-RPC re…

YarsinCitrix:
…sponse

The previous changes for CA-236855 no longer tolerant junk at the end of
input from the pvsproxy, which makes the interface more fragile. To build
a more robust interface in case of issue on the pvsproxy side, new option
strict from jsonrpc is used to interpret the response, so tailing junk
is ignored.

### DIFF
--- a/test/test_jsonrpc_client.ml
+++ b/test/test_jsonrpc_client.ml
@@ -42,7 +42,7 @@ module Input_json_object = Generic.Make (struct
 		let response =
 			try
 				let json = Jsonrpc_client.timeout_read (Unix.descr_of_in_channel fin) 5_000_000_000L in
-				let rpc = Jsonrpc.of_string json in
+				let rpc = Jsonrpc.of_string ~strict:false json in
 				Right rpc
 			with
 			| End_of_file -> Left End_of_file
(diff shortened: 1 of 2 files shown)


YarsinCitrix:
What shall I do to have Travis use the new xs-opam 4.5.0?

mseri:
Has it been released in XenServer-specs?

YarsinCitrix:
no, it is in ring3/master, not in master yet. Does the build scripts only pull from master?

mseri:
If the build is green (including bbt) it should be fine. Ask @gabori to retrigger the upload script in case

test/test_jsonrpc_client.ml
```diff
@@ -59,6 +59,9 @@ module Input_json_object = Generic.Make (struct
 		(* A file containing a partial JSON object. *)
 		"short_call.json", Left Parse_error;
 
+		(* A file containing a JSON object, plus some more characters at the end. *)
+		"good_call_plus.json", Right good_call;
```
robhoes: Did you need to `git add` this file?

YarsinCitrix: this is an existing file, which is not removed the last time we remove this case.

PR #144, opened 2018-06-26
network_monitor_thread: use Xmlrpc_client instead of Rpc_client

edwintorok:
For compatibility with new rpclib 5.8.0, see failure in https://github.com/xapi-project/xs-opam/pull/291 -
the Rpc_client module has been removed from ocaml-rpc

### DIFF
--- a/networkd/network_monitor_thread.ml
+++ b/networkd/network_monitor_thread.ml
@@ -28,11 +28,9 @@ let monitor_whitelist = ref [
 	"vif"; (* This includes "tap" owing to the use of standardise_name below *)
 ]
 
-let xapi_rpc request =
-	Rpc_client.do_rpc_unix
-		~content_type:(Rpc_client.content_type_of_string "text/xml")
-		~filename:(Filename.concat "/var/lib/xcp" "xapi")
-		~path:"/" request
+let xapi_rpc xml =
+  let open Xmlrpc_client in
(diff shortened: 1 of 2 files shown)


edwintorok:
Lets wait for a travis build

networkd/network_monitor_thread.ml
```diff
@@ -28,11 +28,9 @@ let monitor_whitelist = ref [
...
 ]
 
-let xapi_rpc request =
-	Rpc_client.do_rpc_unix
-		~content_type:(Rpc_client.content_type_of_string "text/xml")
-		~filename:(Filename.concat "/var/lib/xcp" "xapi")
-		~path:"/" request
+let xapi_rpc xml =
+  let open Xmlrpc_client in
+  XMLRPC_protocol.rpc ~srcstr:"v6" ~dststr:"xapi" ~transport:(Unix "/var/xapi/xapi") ~http:(xmlrpc ~version:"1.0" "/") xml
```
edwintorok: srcstr should not be `v6`, please use something like `xcp-networkd`

edwintorok: Minor: `/var/xapi` is a symlink to `/var/lib/xcp`, they both work but I think the former is deprecated, so we could use the `/var/lib/xcp/xapi` path that networkd used before, but either way should work.

gaborigloi:
I've squashed the commits

gaborigloi:
There is a BVT in progress.

robhoes:
A good test for this would be to have a PIF with DHCP configuration, and checking that the `PIF.IP` field gets set when the DHCP address comes in.

gaborigloi:
I've tested this by adding a new DHCP network in XenCenter and the correct IP appeared soon afterwards.

PR #146, opened 2018-10-02
More log-related cleanup

robhoes:
Depends on https://github.com/xapi-project/xcp-idl/pull/249.


lib/network_utils.ml
```diff
@@ -905,6 +905,27 @@ module Proc = struct
 end
 
 module Ovs = struct
+  let match_multiple patterns s =
```
lindig: I think it would be good to compile the patterns once when the module is initialised:

```
let patterns = List.map Re.Perl.compile_pat
  [ "no bridge named (.*)\n"
  ; "no row \"(.*)\" in table Bridge"
  ]
```

Afterwards, only use `patterns` for matching.

lib/network_utils.ml
```diff
@@ -60,7 +60,22 @@ let igmp_query_maxresp_time = ref "5000"
 let enable_ipv6_mcast_snooping = ref false
 let mcast_snooping_disable_flood_unregistered = ref true
 
-let check_n_run run_func script args =
+let default_error_handler script args stdout stderr exn =
+  let message =
+    match exn with
+    | Unix.WEXITED n -> Printf.sprintf "Exit code %d" n
+    | Unix.WSIGNALED s -> Printf.sprintf "Signaled %d" s (* Note that this is the internal ocaml signal number, see Sys module *)
+    | Unix.WSTOPPED s -> Printf.sprintf "Stopped %d" s
```
lindig: Should this include a wildcard for other exceptions?

robhoes: Hmm... this is not actually an exception, but a value of type `Unix.process_status`, which only has these three constructors. I'll clarify the names.

robhoes:
@lindig Thanks, I've updated those two.

lib/network_utils.ml
```diff
@@ -905,6 +905,27 @@ module Proc = struct
...
+    | [] -> None
+    | pattern :: rest ->
+      match Re.exec_opt pattern s with
+      | Some groups -> Some groups
+      | None -> loop rest
+    in
+    loop patterns
+
+  let error_handler script args stdout stderr exn =
+    let patterns = List.map Re.Perl.compile_pat [
```
lindig: This compiles the patterns for every invocation of `error_handler` unless the compiler moves this code out. I think it is safer to have this at the top level of the enclosing module.

robhoes:
Note: this needs to be merged with https://github.com/xapi-project/xcp-idl/pull/249

PR #151, opened 2018-12-05
CP-27898 port to Dune, remove warnings, fix Travis

lindig:
This series of commits cleans up the repository:

* port to Dune
* remove warnings
* update Travis
* update documentation
* add PROFILE=gprof
* update Opam files
* no functional changes


INSTALL
```diff
@@ -3,18 +3,4 @@
 
 The easiest way to install is via opam:
 
-    opam init
-    opam remote add xen-org git://github.com/xen-org/opam-repo-dev
```
edwintorok: Should we link to https://github.com/xapi-project/xs-opam#adding-the-opam-repository?

Makefile
```diff
@@ -1,14 +1,17 @@
 BINDIR ?= /usr/bin
 SBINDIR ?= /usr/sbin
 MANDIR ?= /usr/share/man/man1
+JOBS = $(shell getconf _NPROCESSORS_ONLN)
```
edwintorok: I checked, and dune is pretty good at doing this on its own now since version 1.0.0: https://github.com/ocaml/dune/blob/master/CHANGES.md#100-10072018
So I think we can drop the -j and JOBS flag now that we switched to `dune`.

lib/network_utils.ml
```diff
@@ -748,7 +747,7 @@ module Dhclient = struct
     let ipv6' = if ipv6 then "6" else "" in
     Filename.concat "/var/lib/xcp" (Printf.sprintf "dhclient%s-%s.conf" ipv6' interface)
 
-  let generate_conf ?(ipv6=false) interface options =
+  let[@warning "-27"] generate_conf ?(ipv6=false) interface options =
```
edwintorok: I assume `ipv6` was the unused one here, I don't really have a nice solution for that, so lets keep it as is.

PR #154, opened 2019-02-12
CA-310516: Detect vlan created on first boot as managed

psafont:
On first boot networkd may create a vlan named xentemp and the firstboot script call xapi to create that same vlan to get recognized by xapi.
networkd need to recognise this case, otherwise an error occurs and the setup fails.

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -249,7 +249,7 @@ module Sysfs = struct
                  |> (fun p -> try read_one_line p |> duplex_of_string with _ -> Duplex_unknown)
     in (speed, duplex)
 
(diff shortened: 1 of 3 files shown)


lib/network_utils.ml
```diff
@@ -51,6 +51,7 @@ let bonding_dir = "/proc/net/bonding/"
 let uname = ref "/usr/bin/uname"
 let dracut = ref "/sbin/dracut"
 let modinfo = ref "/sbin/modinfo"
+let temp_vlan = "xentemp"
```
robhoes: That's not the best place. This module is for interfacing with system networking, and this doesn't belong in that category. I would just put it in network_config.ml.

PR #155, opened 2019-02-13
CA-310413: Unconditionally apply enic workaround (again)

robhoes:
This reverts 9d5309e8018664ed2766e52aeeb1f39d26c5a88d. The reason is that the
workaround is once again needed on the current version of the enic driver.


thomasmck:
What testing is needed to identify this? Deli made the change as part of https://issues.citrite.net/browse/CP-30068 and run multiple tests at the time which passed.

PR #156, opened 2019-02-16
Enable IPv6 on interfaces when enabled globally and configured

mas90:
By default XenServer disables IPv6 globally in `/etc/sysconfig/network` and via the sysctl `net.ipv6.conf.{default,all}.disable_ipv6=1`.  `xe-enable-ipv6` changes `/etc/sysconfig/network` but leaves the sysctl alone, for good reason since enabling IPv6 on every interface would cause dom0 to configure link-local addresses on every bridge.  Currently, xcp-networkd refuses to configure IPv6 on an interface when sysctl `net.ipv6.conf.all.disable_ipv6` is set.

Instead, distinguish between the two and have xcp-networkd enable IPv6 on an interface when `xe-enable-ipv6` has been used and an IPv6 configuration mode has been set on that interface.

This has no effect by default on XenServer, unless the user has run `xe-enable-ipv6` or otherwise manually enabled it system-wide (or where `/etc/sysconfig/network` doesn't exist).

Also, correct a couple of cases where IPv6 users might end up with no DNS or no link-local address.

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -821,6 +821,15 @@ module Sysctl = struct
   let write value variable =
     ignore (call_script sysctl ["-q"; "-w"; variable ^ "=" ^ value])
 
+  let set_ipv6_enabled interface value =
+    try
+      let variable = "net.ipv6.conf." ^ interface ^ ".disable_ipv6" in
+      let value' = if value then "0" else "1" in
+      write value' variable
+    with
+    | e when value = true -> raise e
+    | _ -> ()
+
   let set_ipv6_autoconf interface value =
     try
       let variables = [
@@ -885,12 +894,17 @@ module Proc = struct
       error "Error: could not read /proc/net/vlan/config";
       []
 
-  let get_ipv6_disabled () =
+  let get_ipv6_default_disabled () =
     try
-      Unixext.string_of_file "/proc/sys/net/ipv6/conf/all/disable_ipv6"
+      Unixext.string_of_file "/proc/sys/net/ipv6/conf/default/disable_ipv6"
       |> String.trim
       |> (=) "1"
     with _ -> false
+
+  let get_ipv6_globally_disabled () =
+    let file = String.trim (try Unixext.string_of_file "/etc/sysconfig/network" with _ -> "") in
+    let lines = List.map String.trim (Astring.String.cuts ~empty:false ~sep:"\n" file) in
+    List.mem "NETWORKING_IPV6=NO" lines
 end
 
 module Ovs = struct
--- a/networkd/network_server.ml
+++ b/networkd/network_server.ml
@@ -533,10 +533,14 @@ module Interface = struct
                                      ethtool_settings; ethtool_offload; _} as c)) ->
             update_config name c;
             exec (fun () ->
-                (* We only apply the DNS settings when in static IPv4 mode to avoid conflicts with DHCP mode.
+                (* We only apply the DNS settings when not in a DHCP mode to avoid conflicts.
                    					 * The `dns` field should really be an option type so that we don't have to derive the intention
                    					 * of the caller by looking at other fields. *)
-                match ipv4_conf with Static4 _ -> set_dns () dbg ~name ~nameservers ~domains | _ -> ());
+                match (ipv4_conf, ipv6_conf) with
+                  | (Static4 _, _)
+                  | (_, Static6 _)
+                  | (_, Autoconf6) -> set_dns () dbg ~name ~nameservers ~domains
+                  | _ -> ());
             exec (fun () -> set_ipv4_conf dbg name ipv4_conf);
             exec (fun () -> match ipv4_gateway with None -> () | Some gateway ->
                 set_ipv4_gateway () dbg ~name ~address:gateway);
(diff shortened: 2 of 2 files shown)


lib/network_utils.ml
```diff
@@ -821,6 +821,15 @@ module Sysctl = struct
   let write value variable =
     ignore (call_script sysctl ["-q"; "-w"; variable ^ "=" ^ value])
 
+  let set_ipv6_enabled interface value =
```
lindig: Minor suggestions:
* use `enabled` instead of `value`
* simplify `when value = true` to `when enabled`
* use `printf` to build the string for `variable`
* is `if value then "0" else "1"` correct? It writes `0` for `false`

robhoes: I agree with the above, except for the last point.

lindig: Sorry, it writes `0` for `true`, which is counter-intuitive.

mas90: Yes, the sysctl is "disable_ipv6", and it is (on XenServer) set to 1 by default.  To enable IPv6, you set disable_ipv6=0.  I could have named the function "set_ipv6_disabled" and had the sense match that of the sysctl, but I thought that was more confusing!

I'll try to make the code a bit clearer.

networkd/network_server.ml
```diff
@@ -314,52 +315,63 @@ module Interface = struct
...
         | Linklocal6 ->
           if List.mem name (Sysfs.list ()) then begin
+            (* xapi sets Linklocal6 if IPv6 has not been configured;
+             * if IPv6 is not enabled globally, leave it disabled *)
             if Dhclient.is_running ~ipv6:true name then
               ignore (Dhclient.stop ~ipv6:true name);
-            Sysctl.set_ipv6_autoconf name false;
-            Ip.flush_ip_addr ~ipv6:true name;
-            Ip.set_ipv6_link_local_addr name
+            if Proc.get_ipv6_default_disabled () then
```
robhoes: I don't really get why we need to distinguish between "default" and "globally" disabled. Where is the thing that would actually set `/proc/sys/net/ipv6/conf/default/disable_ipv6` to `0`?

When will we actually end up in the "true" branch below, for "globally enabled" plus "default disabled"?

robhoes:
I'm not convinced that this is the fix that we need, or that the `xe-enable-ipv6` is just wrong now and needs to be updated. The file I don't think that the file /etc/sysconfig/network is actually used, since it is xcp-networkd that configures the interfaces based on config from xapi.

Should we instead update `xe-enable-ipv6` to update `/etc/sysctl.d/90-net.conf`? It currently has the following hardcoded:
```
# Disable IPv6 on interfaces
net.ipv6.conf.all.disable_ipv6 = 1
net.ipv6.conf.default.disable_ipv6 = 1
```

robhoes:
I'll close this for the moment, though please reopen if there is any news.

olivierlambert:
Hi @robhoes 

We might put some resources to add IPv6 support in XCP-ng. I suppose this is the first place to try? Do you have suggestion on what to do first? Should we go into a different direction or continuing the work started here?

robhoes:
@olivierlambert We'll probably need something like this plus an update to the `xe-enable-ipv6` script (in xen-api/scripts). At least, it wasn't clear to me how this PR alone is enough (though I haven't looked very carefully).

olivierlambert:
Understoond. Pinging @benjamreis as he'll be in charge of this :+1:

olivierlambert:
Also ping @mas90 we'll contact you somehow to discuss together your PR, helping us to get to the end result faster :+1:

benjamreis:
So, I've been playing around with the IPv6 settings on a host and I'll post here what I found, it might be obvious to some but it might be useful for others.
- `xe-enable-ipv6` modify `/etc/sysconfig/network`'s `NETWORKING_IPV6` entry
- `/proc/sys/net/ipv6/conf/{default/al}/disable_ipv6` takes its value from `/etc/sysctl.d/90-net.conf` corresponding entry

From my testing it seems only `/etc/sysctl.d/90-net.conf` has an impact on wether the host can have IPv6 addresses or not. So I wonder what's `xe-enable-ipv6` is used for.

If I understood correctly @robhoes suggested to modify `xe-enable-ipv6` to modify `/etc/sysctl.d/90-net.conf` as well. Am I correct?

@mas90 since your the original author of this PR, would you be comfortable with that?

I think this PR contains necessary fix regardless of the POV about `xe-enable-ipv6`

networkd/network_server.ml
```diff
@@ -521,10 +533,14 @@ module Interface = struct
...
-                (* We only apply the DNS settings when in static IPv4 mode to avoid conflicts with DHCP mode.
+                (* We only apply the DNS settings when not in a DHCP mode to avoid conflicts.
                    					 * The `dns` field should really be an option type so that we don't have to derive the intention
                    					 * of the caller by looking at other fields. *)
-                match ipv4_conf with Static4 _ -> set_dns () dbg ~name ~nameservers ~domains | _ -> ());
+                match (ipv4_conf, ipv6_conf) with
+                  | (Static4 _, _)
+                  | (_, Static6 _)
+                  | (_, Autoconf6) -> set_dns () dbg ~name ~nameservers ~domains
+                  | _ -> ());
```
benjamreis, line 543: This fix is available alone in a new PR (https://github.com/xapi-project/xcp-networkd/pull/181) since the rest of this one is being discussed.

PR #157, opened 2019-02-21
CA-311211: Fix destroy_existing_vlan_bridge when enic workaround is e…

robhoes:
…nabled

The function `destroy_existing_vlan_bridge` is called from `Bridge.create` if
the bridge to be created is for a VLAN. The purpose is to get rid of any
existing bridge for the same VLAN but with a different name. This was not
working properly if case the enic workaround is in place, and the Linux bridge
backend is used.

The main use of this is installing a host with the management interface on a
VLAN, where early in the host boot process, xcp-networkd creates a temporary
bridge for the VLAN, called `xentemp`. This is because xcp-networkd does not
yet know the final name of the VLAN bridge, which is determined later by xapi.
When firstboot scripts run, a VLAN is created through xapi, with a new bridge
name that is meant to replace the `xentemp` bridge.

For example, for VLAN 123, the output of `brctl show` would be like this after
xcp-network has started:

    bridge name        bridge id                STP enabled       interfaces
    xenbr0             8000.0cc47ae6872c        no                eth0
    xentemp            8000.0cc47ae6872d        no                eth0.123

The desired state after the firstboot scripts run is:

    bridge name        bridge id                STP enabled       interfaces
    xenbr0             8000.0cc47ae6872c        no                eth0
    xapi0              8000.0cc47ae6872d        no                eth0.123

Now, the problem occurred when running on a Cisco UCS host, with the "enic
workaround" enabled. What happens then, is that not the eth0, but a VLAN0
device eth0.0 is added to the "parent" bridge xenbr0:

    bridge name        bridge id                STP enabled       interfaces
    xenbr0             8000.0cc47ae6872c        no                eth0.0
    xentemp            8000.0cc47ae6872d        no                eth0.123

The function `destroy_existing_vlan_bridge` did not take this into account. It
is fixed by push the call to this function further down into `Bridge.create`,
below the point where the enic workaround has been dealt with. It turned out
easier to split the function into separate OVS and Linux bridge versions.

### DIFF
--- a/networkd/network_server.ml
+++ b/networkd/network_server.ml
@@ -572,48 +572,43 @@ module Bridge = struct
         | Bridge -> Sysfs.get_all_bridges ()
       ) ()
 
-  let destroy_existing_vlan_bridge name (parent, vlan) =
-    begin match !backend_kind with
-      | Openvswitch ->
-        let bridges =
-          let raw = Ovs.vsctl ["--bare"; "-f"; "table"; "--"; "--columns=name"; "find"; "port"; "fake_bridge=true"; "tag=" ^ (string_of_int vlan)] in
-          if raw <> "" then Astring.String.cuts ~empty:false ~sep:"\n" (String.trim raw) else []
-        in
-        let existing_bridges =
-          List.filter ( fun bridge ->
-              match Ovs.bridge_to_vlan bridge with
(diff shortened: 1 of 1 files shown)


robhoes:
Hint: it is best to review the diff with whitespace suppressed.

PR #158, opened 2019-04-03
Backport CP-22034 to lcm branch

psafont:
The last commit looks messy, but most of it is generated code.

`lib/link_stubs.c` is removed and 2 lines from `_oasis` are removed so the build system does not try to compile it. The rest of the changes are due to running `oasis setup`

Analogous PR for dundee: https://github.com/xenserver/xcp-networkd/pull/15


edwintorok:
The code looks good. Travis has some problems on this branch, were you able to build this locally?

psafont:
@edwintorok the code is buildable

the travis failures seem to happen because the xs-opam version is much newer than the one needed for an lcm build (it's using 6.5.0)

PR #159, opened 2019-06-05
CA-320563: xcp-networkd: fix the regex match for PCI BDF string

minglumlu:
PCI BDF format: up to 256 buses, each with up to 32 devices, each
supporting eight functions.

This commit fixes the regex match to include hex strings.

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -246,7 +246,7 @@ module Sysfs = struct
     try
       Sys.readdir ("/sys/bus/pci/drivers/" ^ driver)
       |> Array.to_list
-      |> List.filter (Re.execp (Re.Perl.compile_pat "\\d+:\\d+:\\d+\\.\\d+"))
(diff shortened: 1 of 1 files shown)


lib/network_utils.ml
```diff
@@ -246,7 +246,7 @@ module Sysfs = struct
     try
       Sys.readdir ("/sys/bus/pci/drivers/" ^ driver)
       |> Array.to_list
-      |> List.filter (Re.execp (Re.Perl.compile_pat "\\d+:\\d+:\\d+\\.\\d+"))
+      |> List.filter (Re.execp (Re.Perl.compile_pat "\\d+:[a-f\\d]+:[a-f\\d]+\\.\\d+"))
```
lindig: What about uppercase hex letters `A` to `F`? To avoid confusion with backslah-escaping, this could be also written as:

```
{|\d+:[a-f\d]+:[a-f\d]+\.\d|}
```

minglumlu: Hi @lindig 
It's good to know this kind of regex string. Thanks!
And for the upper cases, the name of files under `/sys/bus/pci/drivers/` are always in lower cases. So I think it is not necessary to add matching for A-F.

PR #160, opened 2019-06-11
Restart 'ip monitor address' process if it quitted

cheese:
It the 'ip monitor address' process quitted for some reason, xcp-networkd
will be logging in an infinite loop and soon eat up the /var/log space.

### DIFF
--- a/networkd/network_monitor_thread.ml
+++ b/networkd/network_monitor_thread.ml
@@ -278,7 +278,7 @@ let clear_input fd =
   loop ();
   Unix.clear_nonblock fd
 
-let ip_watcher () =
+let rec ip_watcher () =
   let cmd = Network_utils.iproute2 in
   let args = ["monitor"; "address"] in
   let readme, writeme = Unix.pipe () in
@@ -299,12 +299,26 @@ let ip_watcher () =
     end;
     loop ()
   in
+  let restart_ip_watcher () = begin
+    Unix.close readme;
+    Thread.delay 5.0;
+    ip_watcher ();
+  end
+  in
   while true do
     try
       info "(Re)started IP watcher thread";
       loop ()
     with e ->
-      warn "Error in IP watcher: %s\n%s" (Printexc.to_string e) (Printexc.get_backtrace ())
+      warn "Error in IP watcher: %s\n%s" (Printexc.to_string e) (Printexc.get_backtrace ());
+      match !watcher_pid with
+      | None -> restart_ip_watcher ()
+      | Some pid ->
+        let quitted, _ = Forkhelpers.waitpid_nohang pid in
+        if quitted <> 0 then begin
+          warn "address monitoring process quitted, try to restart it";
+          restart_ip_watcher ()
+        end
   done
 
 let start () =


lindig:
Thank you for your PR.  

* I suspect the Travis build will fail; to fix it use `DISTRO="debian-unstable"` in `.travis.yml`
* please sign your commit (`Signed-off-by:...`) by using `git commit -s`

lindig:
* Would it be better to let systemd control the restarts by terminating xcp-systemd? It can limit the number of restarts if they fail quickly.
* Should restarts be delayed by a few seconds to avoid tight loops in case of repeated failures?

cheese:
>     * Would it be better to let systemd control the restarts by terminating xcp-systemd? It can limit the number of restarts if they fail quickly.
xcp-networkd itself won't terminate or restart if 'ip monitor address' quitted.

> 
>     * Should restarts be delayed by a few seconds to avoid tight loops in case of repeated failures?
That's a good idea. Sleep for 5 seconds?

lindig:
That was my question: instead of calling `ip_watcher` recursively, should we terminate xcp-networkd instead? A `Thread.delay 5.0` seems like a good idea to me.

robhoes:
xcp-networkd does a lot more than running ip-watcher thread, so we should not terminate it.

cheese:
Delay added at last update

networkd/network_monitor_thread.ml
```diff
@@ -299,13 +299,28 @@ let ip_watcher () =
...
-      warn "Error in IP watcher: %s\n%s" (Printexc.to_string e) (Printexc.get_backtrace ())
-  done
+  try
+    while true do
+      try
+        info "(Re)started IP watcher thread";
+        loop ()
+      with e -> begin
+        warn "Error in IP watcher: %s\n%s" (Printexc.to_string e) (Printexc.get_backtrace ());
+        Thread.delay 5.0;
```
lindig: I think this is not the best place because now we will introduce a gap in the log file. I think it is better to do this at the end of `with e -> begin ...; Thread.delay 5.0; end`.

cheese: So move the delay right before next recursive call.

networkd/network_monitor_thread.ml
```diff
@@ -299,13 +299,28 @@ let ip_watcher () =
...
     loop ()
   in
-  while true do
-    try
-      info "(Re)started IP watcher thread";
-      loop ()
-    with e ->
-      warn "Error in IP watcher: %s\n%s" (Printexc.to_string e) (Printexc.get_backtrace ())
-  done
+  try
```
robhoes: I think that the idea is good in general, but the function is getting a little confusing now, with the infinite loop as well as the recursive call, and the double exception handler. I think we just need to remove the while loop and collapse the exception handlers.

networkd/network_monitor_thread.ml
```diff
@@ -299,13 +299,28 @@ let ip_watcher () =
...
+    while true do
+      try
+        info "(Re)started IP watcher thread";
+        loop ()
+      with e -> begin
```
robhoes: The `begin` and `end` are not needed here.

networkd/network_monitor_thread.ml
```diff
@@ -299,13 +299,28 @@ let ip_watcher () =
...
+        | None -> raise e
+        | Some pid ->
+          let quited, _ = Forkhelpers.waitpid_nohang pid in
+          if quited <> 0 then begin
+            warn "address monitoring process quitted, try to restart it";
+            raise e;
+          end;
+      end;
+    done
+  with _ -> begin
```
robhoes: The `begin` and `end` are not needed here.

networkd/network_monitor_thread.ml
```diff
@@ -299,13 +299,28 @@ let ip_watcher () =
...
+        loop ()
+      with e -> begin
+        warn "Error in IP watcher: %s\n%s" (Printexc.to_string e) (Printexc.get_backtrace ());
+        match !watcher_pid with
+        | None -> raise e
+        | Some pid ->
+          let quited, _ = Forkhelpers.waitpid_nohang pid in
+          if quited <> 0 then begin
+            warn "address monitoring process quitted, try to restart it";
+            raise e;
```
robhoes: In OCaml, semicolons are used in between sequential statements, and are not needed after the last one in a block. I.e this line, and the two lines below, do not need semicolons.

networkd/network_monitor_thread.ml
cheese: The while loop seems not feasible to remove. I now removed one level of exception handler.

cheese:
Sorry for forgetting to update the patch.

PR #161, opened 2019-06-14
CA-320563: re-submit for missing change on regex

minglumlu:
PCI BDF format: up to 256 buses, each with up to 32 devices, each
supporting eight functions.

This commit fixes the regex match to include hex strings.

Signed-off-by: Ming Lu <ming.lu@citrix.com>


minglumlu:
Hi @lindig 
I apologize for my mistake that not pushed the rework based on your comment in last PR. I think I missed the push operation in last time, I just left it in my private build at that time.

PR #162, opened 2019-06-25
Extend the naughty DHCP-server workaround to DNS

robhoes:
See aead1477.

### DIFF
--- a/networkd/network_server.ml
+++ b/networkd/network_server.ml
@@ -250,15 +250,7 @@ module Interface = struct
         | DHCP4 ->
           let open Xapi_stdext_monadic in
           let gateway = Opt.default [] (Opt.map (fun n -> [`gateway n]) !config.gateway_interface) in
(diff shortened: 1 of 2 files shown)


networkd/network_server.ml
```diff
@@ -250,15 +250,7 @@ module Interface = struct
         | DHCP4 ->
```
edwintorok: How about DHCPv6?

edwintorok:
The IPv4 changes look correct, what do we need to do about IPv6? (also a ticket reference would be useful in case we want to backport)

robhoes:
The `Dhclient.start` function being updated here is also used for DHCPv6, so that those calls also get `PEERDNS=no`. Generally speaking, DNS setting currently goes through the v4 path only; it's a host setting and should be set only once if an interface has both v4 and v6. It's a bit messy... :(

PR #163, opened 2019-07-16
CA-323523: Ensure that dhclients get restarted when gateway/DNS changes

robhoes:


edwintorok:
There are a few more places that update the config, such as reset_state/clear_state/update_config/remove_config/etc.
Should these behave the same way? Should there be a function set_config that all these functions call that checks whether the new config and the old config are different, and then remove the config file if so?

robhoes:
@edwintorok That's a different config ;)

lib/network_utils.ml
```diff
@@ -722,7 +722,14 @@ module Linux_bonding = struct
       error "Bond %s does not exist; cannot set properties" master
 end
 
-module Dhclient = struct
+module Dhclient :
+sig
+  val remove_conf_file : ?ipv6:bool -> string -> unit
```
lindig: This (or labels) could add some implicit documentation:
```
type interface = string (** "eth0" *)
val remove_conf_file : ?ipv6:bool -> interface -> unit
...
```

lib/network_utils.ml
```diff
@@ -759,6 +766,12 @@ module Dhclient = struct
     let conf = generate_conf ~ipv6 interface options in
     Xapi_stdext_unix.Unixext.write_string_to_file (conf_file ~ipv6 interface) conf
 
+  let remove_conf_file ?(ipv6=false) interface =
+    let file = conf_file ~ipv6 interface in
+    try
+      Unix.unlink file
+    with _ -> ()
```
lindig: Maybe log this?
```
| exn-> warn "failed to remove file %s: %s" file (Printexc.to_string exn)
 ```

robhoes: If it fails, that's probably because the file doesn't exist, which is fine.

PR #164, opened 2019-07-26
CP-28368 Propagate test_highlevel ounit removal

lippirk:
Signed-off-by: lippirk <ben.anson@citrix.com>

### DIFF
--- a/test/network_test.ml
+++ b/test/network_test.ml
@@ -12,15 +12,9 @@
  * GNU Lesser General Public License for more details.
  *)
 
-open OUnit
-
-let base_suite =
-  "base_suite" >:::
-  [
-    Network_test_lacp_properties.suite;
-    Test_jsonrpc_client.suite;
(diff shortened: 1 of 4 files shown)


test/network_test.ml
```diff
@@ -12,15 +12,9 @@
...
-
-let base_suite =
-  "base_suite" >:::
-  [
-    Network_test_lacp_properties.suite;
-    Test_jsonrpc_client.suite;
-  ]
-
-let _ = 
-  Coverage.init "network_test";
```
robhoes: Was this not doing anything?

robhoes: @edwintorok says it may not actually work...

lippirk: i shouldn't have removed it, my bad

it sets the `BISECT_FILE` env variable - i will check that it is still behaving reasonably

PR #167, opened 2019-08-13
CA-323765: preserve the "type" config when re-adding existing interfaces

zli:
... such as those pvsproxy ones which have type=internal.

Signed-off-by: Zheng Li <zheng.li3@citrix.com>

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -1127,6 +1127,12 @@ module Ovs = struct
         fork_script !inject_igmp_query_script (["--no-check-snooping-toggle"; "--max-resp-time"; !igmp_query_maxresp_time] @ bvifs' @ vvifs)
       with _ -> ()
 
+    let create_port_arg ?ty name bridge =
+      let type_args = match ty with
+        | None | Some "" -> []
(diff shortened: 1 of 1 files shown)


lib/network_utils.ml
```diff
@@ -1162,7 +1168,15 @@ module Ovs = struct
       in
       let vif_arg =
         let existing_vifs = List.filter (fun iface -> not (Sysfs.is_physical iface)) (bridge_to_interfaces name) in
-        List.flatten (List.map (fun vif -> ["--"; "--may-exist"; "add-port"; name; vif]) existing_vifs)
+        let ifaces_with_type =
+          let raw = vsctl ~log:true ["--bare"; "-f"; "table"; "--"; "--columns=name,type"; "find"; "interface"; "type!=\\\"\\\""] in
```
edwintorok: If you want to avoid all the  extra quoting and slashes you can use `{|type!=\"\"|}` instead, it is equivalent to `"type!=\\\"\\\""`

zli: Interesting, learnt something new. Just update.

edwintorok: It got introduced in OCaml 4.02: http://caml.inria.fr/pub/docs/manual-ocaml/manual036.html

PR #171, opened 2019-11-14
CA-329442: Avoid recreating bridges unless absolutely necessary

robhoes:
Fixes XSI-418.

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -1179,8 +1179,16 @@ module Ovs = struct
         List.flatten (List.map (fun vif -> create_port_arg ?ty:(List.assoc_opt vif ifaces_with_type) vif name) existing_vifs)
       in
       let del_old_arg =
-        if vlan <> None then
-          (* This is to handle the case that a "real" bridge (not a "fake" VLAN bridge) already exists *)
+        let real_bridge_exists () =
+          try
+            (* `ovs-vsctl br-to-parent <name>` returns <name> if <name> is a current "real" bridge *)
+            vsctl ~log:false ["br-to-parent"; name] |> String.trim = name
+          with _ -> false
+        in
+        if vlan <> None && real_bridge_exists () then
+          (* This is to handle the case that a "real" bridge (not a "fake" VLAN bridge)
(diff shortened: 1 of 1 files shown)


lib/network_utils.ml
```diff
@@ -1204,8 +1212,20 @@ module Ovs = struct
           ["--"; "set"; "bridge"; name; "other_config:mcast-snooping-disable-flood-unregistered=" ^ (string_of_bool !mcast_snooping_disable_flood_unregistered)]
         | _ -> []
       in
-      vsctl (del_old_arg @ ["--"; "--may-exist"; "add-br"; name] @
-                       vlan_arg @ mac_arg @ fail_mode_arg @ disable_in_band_arg @ external_id_arg @ vif_arg @ set_mac_table_size @ set_igmp_snooping @ set_ipv6_igmp_snooping @ disable_flood_unregistered)
```
edwintorok: This is just a formatting change right?

lib/network_utils.ml
```diff
@@ -1179,8 +1179,16 @@ module Ovs = struct
         List.flatten (List.map (fun vif -> create_port_arg ?ty:(List.assoc_opt vif ifaces_with_type) vif name) existing_vifs)
       in
       let del_old_arg =
-        if vlan <> None then
-          (* This is to handle the case that a "real" bridge (not a "fake" VLAN bridge) already exists *)
+        let real_bridge_exists () =
+          try
+            (* `ovs-vsctl br-to-vlan <name>` returns 0 if <name> is a current "real" bridge *)
+            vsctl ~log:false ["br-to-vlan"; name] |> String.trim = "0"
```
edwintorok: Manpage says that it will print `0` for a real bridge (note that this is not an exit-code, it is an output string),
but what about a VLAN with id of 0?

lib/network_utils.ml
robhoes: Yes indeed.

lib/network_utils.ml
robhoes: Good spot. I've switched to `br-to-parent` now, which avoids the problem.

PR #175, opened 2020-04-06
Make dhclient send our hostname to the DHCP server

johnelse:
This generates a config file of the form:

```
interface "xenbr0" {
  send host-name = gethostname();
  request subnet-mask, broadcast-address, time-offset, host-name, nis-domain, nis-servers, ntp-servers, interface-mtu, routers, domain-name, domain-name-servers;
}
```

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -760,7 +760,8 @@ struct
       else (debug "%s is NOT the DNS interface" interface; [])
     in
     let request = minimal @ set_gateway @ set_dns in
-    Printf.sprintf "interface \"%s\" {\n  send %s;\n  request %s;\n}\n" interface send (String.concat ", " request)
+    Printf.sprintf "interface \"%s\" {\n  send %s;\n  request %s;\n}\n"
+      interface send (String.concat ", " request)
 
   let read_conf_file ?(ipv6=false) interface =
     let file = conf_file ~ipv6 interface in
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -746,6 +746,7 @@ struct
     Filename.concat "/var/lib/xcp" (Printf.sprintf "dhclient%s-%s.conf" ipv6' interface)
 
   let[@warning "-27"] generate_conf ?(ipv6=false) interface options =
+    let send = "host-name = gethostname()" in
     let minimal = ["subnet-mask"; "broadcast-address"; "time-offset"; "host-name"; "nis-domain";
                    "nis-servers"; "ntp-servers"; "interface-mtu"] in
     let set_gateway =
@@ -759,7 +760,7 @@ struct
       else (debug "%s is NOT the DNS interface" interface; [])
     in
     let request = minimal @ set_gateway @ set_dns in
-    Printf.sprintf "interface \"%s\" {\n  request %s;\n}\n" interface (String.concat ", " request)
+    Printf.sprintf "interface \"%s\" {\n  send %s;\n  request %s;\n}\n" interface send (String.concat ", " request)
 
   let read_conf_file ?(ipv6=false) interface =
     let file = conf_file ~ipv6 interface in
--- a/test/network_test_lacp_properties.ml
+++ b/test/network_test_lacp_properties.ml
@@ -51,6 +51,7 @@ module OVS_Cli_test = struct
   include Ovs.Cli
   let vsctl_output = ref []
   let vsctl ?log args =
+    (* Ignore unused argument to avoid warning under dune's default settings. *)
     ignore log;
     vsctl_output := args ;
     String.concat " " args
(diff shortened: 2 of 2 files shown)


robhoes:
Hey John, why is this useful?

test/network_test_lacp_properties.ml
```diff
@@ -51,6 +51,7 @@ module OVS_Cli_test = struct
   include Ovs.Cli
   let vsctl_output = ref []
   let vsctl ?log args =
+    ignore(log);
```
robhoes, line 54: Perhaps add a comment here, or someone will delete it :)

lib/network_utils.ml
```diff
@@ -759,7 +760,8 @@ struct
       else (debug "%s is NOT the DNS interface" interface; [])
     in
     let request = minimal @ set_gateway @ set_dns in
-    Printf.sprintf "interface \"%s\" {\n  request %s;\n}\n" interface (String.concat ", " request)
+    Printf.sprintf "interface \"%s\" {\n  send %s;\n  request %s;\n}\n"
```
lindig, line 763: Using `{|...|}` could avoid the dance around escaping quotes.

olivierlambert:
> Hey John, why is this useful?

As soon you enabled DHCP, it's very likely that you want to send the hostname in the request. This allows your DNS server to populate the record at the same time (this is the reason of this feature).

This way, in a DNS/DHCP enabled environment, you don't have to remember the IP address (or even knowing it), you can just rely on the hostname, so you can `ssh` or `ping` it directly with the hostname.

I would say that DHCP without it isn't really interesting. Also, as it's only something used for people who decided willingly to enabled DHCP.

robhoes:
In the DHCP request options we actually request a hostname from the DHCP server, which is then configured on the host together with the IP address etc. So the DHCP server is in charge of hostnames and IP addresses, so that you don't need to manually set a hostname upon installing the host. Isn't that the common use case?

test/network_test_lacp_properties.ml
lindig: Indeed. What does this do and why does it require parentheses?

robhoes:
Anyway, I'm happy to add this, so allow for additional DHCP use cases, as it seems to be complementary to the current config.

olivierlambert:
About "common use case": it's hard to answer :thinking: IMHO, if one decides to enable DHCP, it's specifically to not have to deal with manual configuration in the infrastructure (on the DNS/DHCP server).

I can tell about our own use case: when you use a lot of XCP-ng/CH in a lab in nested to do tons of tests, it's really useful to not have to deal with fixed IP (therefore using DHCP) but also not to deal with tons of "ephemeral" IPs to memorize (you quickly remove/recreate hosts).

What's create with XCP-ng/CH installer: it generates a random host name, so you don't even have to select one manually. This will get rid of having to look at the IP address. Which is worth such a small modification. But again, it's just my opinion and I'm open to suggestions :+1:

robhoes:
I'd probably set up the DHCP server to return random hostnames (e.g. a name plus an incremental number), but as I said, I'm happy to add this to allow for additional use cases.

test/network_test_lacp_properties.ml
johnelse: Good point on the parentheses - I've obviously not been writing enough OCaml recently :) I'll fix that.

Without this, the tests don't build - dune complains about the unused `log` argument. The argument can't go away entirely though, or the `OVS_Cli_test` module won't conform to the `Ovs.Cli` interface.

lindig: Use:
```
let vsctl ?log:_ args =
```

lib/network_utils.ml
johnelse: I've pushed a commit that does this - it requires some odd indentation to get the whitespace right around the newline though.

nagilum99:
Whatever networks I've touched: We never had a DHCP giving names. It's common that e.g. Windows Clients have their name and tell DHCP, which then adds it to the (local) DNS server.

lib/network_utils.ml
johnelse: n.b. the quote escaping was there before my change...but happy to keep the string syntax change or not, whichever you prefer :)

lindig: I know. I'm fine with the way it is now. Neither one is idea.

test/network_test_lacp_properties.ml
```diff
@@ -51,6 +51,8 @@ module OVS_Cli_test = struct
   include Ovs.Cli
   let vsctl_output = ref []
   let vsctl ?log args =
+    (* Ignore unused argument to avoid warning under dune's default settings. *)
+    ignore log;
```
lindig, line 55: Can we get rid of this by using `let vsctl ?log:_ args =`?

lib/network_utils.ml
johnelse: I've reverted this back to the old quote escaping syntax.

test/network_test_lacp_properties.ml
johnelse: Done :)

robhoes:
I've done a squash-and-merge, combining the commit messages a little.

PR #176, opened 2020-05-15
Format ocaml code using ocamlformat

psafont:
This is using the same configuration as https://github.com/xapi-project/xcp-idl/pull/310


psafont:
Now `make format` is phony.

PR #177, opened 2020-06-12
allow use of already enabled sr-iov virtual functions in xapi managed sr-iov networks

robhoes:


robhoes:
This is simply a rebase of #173 .

PR #179, opened 2020-09-25
CP-34942: update for rpclib 8 compatibility

psafont:
Depends on https://github.com/xapi-project/xs-opam/pull/511


lippirk (approved):

edwintorok (approved):
What is all the `is_notification stuff`? Is there a helper function we should be using for building these RPCs that would hide this?

psafont:
Notifications are message-switch calls that don't require a response.

PR #182, opened 2021-01-22
[WIP] Fix dhcp6

benjamreis:
An host doesn't receive an IPv6 address when in DHCP.

### DIFF
--- a/lib/network_utils.ml
+++ b/lib/network_utils.ml
@@ -888,18 +888,32 @@ end = struct
       (Printf.sprintf "dhclient%s-%s.conf" ipv6' interface)
 
   let[@warning "-27"] generate_conf ?(ipv6 = false) interface options =
-    let send = "host-name = gethostname()" in
+    let send = "fqdn.fqdn = gethostname()" in
     let minimal =
       [
         "subnet-mask"
       ; "broadcast-address"
       ; "time-offset"
       ; "host-name"
-      ; "nis-domain"
-      ; "nis-servers"
-      ; "ntp-servers"
       ; "interface-mtu"
       ]
+    in
+    let family_minimal =
+      if ipv6 then
+        [
+          "dhcp6.nis-servers"
+        ; "dhcp6.nisp-servers"
+        ; "nis-domain-name"
+        ; "nisp-domain-name"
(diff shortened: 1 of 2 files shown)


benjamreis:
The address is still not received despite my fixes, I'm still looking for the cause, any advise would be welcomed. :)

lib/network_utils.ml
```diff
@@ -919,7 +919,13 @@ end = struct
...
       )
     in
-    let request = minimal @ set_gateway @ set_dns in
+    let ipv6_fields =
+      if ipv6 then
+        ["dhcp6.client-id"; "dhcp6.domain-search"; "dhcp6.name-servers"]
+      else
+        []
+    in
+    let request = minimal @ set_gateway @ set_dns @ ipv6_fields in
```
psafont, line 928: It's better to avoid chained appends (`@`) like this and use List.flatten instead. This is because of the depth in recursion is much higher in the chained appends.

```suggestion
    let request = List.flatten [minimal; set_gateway; set_dns; ipv6_fields] in
```

psafont:
Is this still WIP?

benjamreis:
Yes, the DHCPv6 still doesn't work: https://github.com/xcp-ng/xcp/issues/475

Investigation is still ongoing.

lib/network_utils.ml
```diff
@@ -913,13 +927,16 @@ end = struct
     let set_dns =
```
lindig, line 927: * `set_dns` is not a good name because this returns a configuration and does not set anyting
* the code could be made flatter which I think is easier to read

```
let dns_config =
  let emit = List.mem (`dns interface) options in
  if emit && ipv6 then
    debug "%s is the DNS interface" interface ;
    ["dhcp6.domain-search"; "dhcp6.name-servers"]
  else if emit && not ipv6 then
    debug "%s is the DNS interface" interface ;
    ["domain-name"; "domain-name-servers"]
  else
    debug "%s is NOT the DNS interface" interface ;
    []
```

lindig (approved):

lindig:
This looks stale. I propose to close this until momentum picks up.

robhoes:
It also needs to be on xen-api now.

Issue #23, opened 2013-10-30
Unguarded List.hd?


djs55:
Against 0.9.3 attempting to bring up a VLAN:

```
Oct 30 14:10:55 brontitall xcp-networkd: [debug|brontitall|271|VLAN.create R:8442d0f031e0|network_server] Bringing up interf
ace brem2
Oct 30 14:10:55 brontitall xcp-networkd: [ info|brontitall|271|VLAN.create R:8442d0f031e0|network_utils] /sbin/ip link set b
rem2 up
Oct 30 14:10:55 brontitall xcp-networkd: [debug|brontitall|271|VLAN.create R:8442d0f031e0|network_server] Adding port em2 to
 bridge brem2 with interfaces em2
Oct 30 14:10:55 brontitall xcp-networkd: [debug|brontitall|271|VLAN.create R:8442d0f031e0|network_server] Bringing up interf
ace em2
Oct 30 14:10:55 brontitall xcp-networkd: [ info|brontitall|271|VLAN.create R:8442d0f031e0|network_utils] /sbin/ip link set e
m2 up
Oct 30 14:10:55 brontitall tapdisk[8275]: tapdisk-control: init, 10 x 4k buffers
Oct 30 14:10:55 brontitall tapdisk[8275]: I/O queue driver: lio
Oct 30 14:10:55 brontitall tapdisk[8275]: tapdisk-log: started, level 0
Oct 30 14:10:55 brontitall xcp-networkd: [debug|brontitall|271|VLAN.create R:8442d0f031e0|network_server] Creating bridge xa
pi1 (VLAN 12 on bridge brem2)
Oct 30 14:10:55 brontitall tapdisk[8275]: nbd: Set up local unix domain socket on path '/var/run/blktap-control/nbdclient827
5'
Oct 30 14:10:55 brontitall xcp-networkd: [ info|brontitall|271|VLAN.create R:8442d0f031e0|network_utils] /sbin/ip link set x
api1 address bc:30:5b:ce:ef:de
Oct 30 14:10:55 brontitall xapi: [debug|brontitall|24752 UNIX /var/lib/xcp/xapi|VLAN.create R:8442d0f031e0|backtrace] Raised
 at file \"ghost-location\", line 1, characters 0-0\nCalled from file \"nm.ml\", line 396, characters 3-54\nCalled from file
 \"network.ml\", line 30, characters 2-6\n
[...]

PR #186, opened 2021-02-16
Remove usages of List.hd

psafont:
List.hd can raise exceptions in locations where they are not expected.
Quite a few deals with string splits, these can be worked around to not
use lists at all.
Others need to raise a semantically-correct error to properly debug the
issue.

Signed-off-by: Pau Ruiz Safont <pau.safont@citrix.com>


lib/network_utils.ml
```diff
@@ -806,16 +807,15 @@ module Linux_bonding = struct
             Sysfs.read_one_line (Sysfs.getpath master ("bonding/" ^ prop))
           in
           if prop = "mode" then
-            Some
-              ( prop
-              , List.hd (Astring.String.cuts ~empty:false ~sep:" " bond_prop) )
+            Astring.String.cut ~sep:" " (String.trim bond_prop)
+            |> Option.map (fun (mode, _) -> (prop, mode))
```
psafont, line 811: Now the debug line won't print if the List.hd fails, I don't think it was meant to be printed.

networkd/network_server.ml
```diff
@@ -903,13 +903,22 @@ module Bridge = struct
...
-                       (fun n ->
-                         Astring.String.is_prefix ~affix:"eth" n
-                         || Astring.String.is_prefix ~affix:"bond" n)
-                       (Sysfs.bridge_to_interfaces parent))
+                  match parent_bridge_interfaces with
+                  | [] ->
+                      error
+                        {|Interface for bridge parent "%s" of vlan %i not found|}
+                        parent vlan ;
+                      raise (Network_error (Bridge_does_not_exist parent))
```
psafont, line 919: Double-check this error makes sense

lib/network_utils.ml
lippirk: I think we only need the `try` around `Sysfs.read_one_line` now

lippirk (approved):

networkd/network_server.ml
```diff
@@ -1160,39 +1170,41 @@ module Bridge = struct
...
+    | Bridge -> (
+      match interfaces with
+      | [iface] ->
+          Interface.bring_up () dbg ~name:iface
+      | _ ->
+          ( Linux_bonding.add_bond_master name ;
+            let bond_properties =
+              if
+                List.mem_assoc "mode" bond_properties
+                && List.assoc "mode" bond_properties = "lacp"
```
lippirk, line 1182:
```suggestion
                List.assoc_opt "mode" bond_properties = Some "lacp"
```

while you are here

lib/network_utils.ml
robhoes: The docs for `cut` say that it returns "None if sep can't be matched in s". So if `bond_prop` does not contain a space, then the result will now be `None`.

For `cuts`, the docs say: "If there is no separator match in s, the list [s] is returned.". So the result would be `Some (prop, bond_prop)`

networkd/network_server.ml
robhoes: It's an interface, so `Interface_does_not_exist`.

robhoes: Well, that also doesn't quite fit, because we don't have a name for that interface :)

robhoes: Perhaps use `Internal_error`, because it is probably a bug if you hit this.

robhoes (changes requested):
This mostly looks good. Two comments above.

lib/network_utils.ml
psafont: The original took the first "word" of the line, ignoring leading spaces.

I think this can be done without using Optional:
```ocaml
let get_mode line =
  let a_space char = char = ' ' in
  Astring.String.(
    line |> drop ~sat:a_space |> take ~sat:(Fun.negate a_space))
in
```

robhoes (approved):

psafont:
I've replaced the travis workflow with a github actions workflow:, which is successful https://github.com/psafont/xcp-networkd/runs/1927777307?check_suite_focus=true

robhoes (approved):

psafont:
https://github.com/xapi-project/xcp-networkd/issues/23#issuecomment-27680200 😀

robhoes:
@psafont it got worse: now I hate _all_ `List.hd`, not just the unguarded ones :)

PR #192, opened 2021-08-20
CP-38064: adapt code to ppxlib 0.9

psafont:
In particular adapts the code to rpclib 7

Created branch 0.56.0-lcm for stockholm

### DIFF
--- a/networkd/network_server.ml
+++ b/networkd/network_server.ml
@@ -1360,7 +1360,11 @@ module PVS_proxy = struct
   let configure_site _dbg config =
     debug "Configuring PVS proxy for site %s" config.site_uuid ;
     let call =
-      {Rpc.name= "configure_site"; params= [Rpcmarshal.marshal t.ty config]}
+      {
+        Rpc.name= "configure_site"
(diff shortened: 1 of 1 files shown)


networkd/network_server.ml
```diff
@@ -1360,7 +1360,11 @@ module PVS_proxy = struct
   let configure_site _dbg config =
     debug "Configuring PVS proxy for site %s" config.site_uuid ;
     let call =
-      {Rpc.name= "configure_site"; params= [Rpcmarshal.marshal t.ty config]}
+      {
+        Rpc.name= "configure_site"
+      ; params= [Rpcmarshal.marshal t.ty config]
+      ; notif= false
```
lindig, line 1366: I think `notfif` as a name is unfortunate. Does this mean "notify" or "not if"? Where is this coming from?

psafont: It means notify, your comments convinced marcello to change it to `notify` in rpclib 8, which is used on master.

I could backport the changes.

lippirk (approved):

lindig (approved):

PR #193, opened 2022-05-12
CA-359978 UPD-825: Flush IP addresses when switching from static to DHCP

lindig:
Backport of be440f369e91fccfc4f94b9ba7ac03f382626067 from xen-api.git

Switching to DHCP does not simply replace existing IP addresses, but
may add one in addition.

Signed-off-by: Christian Lindig <christian.lindig@citrix.com>


psafont (approved):
This is a faithful backport

robhoes (approved):

